PR #5, opened 2017-07-03
Manual reference example

agitter:


references/README.md
```diff
@@ -10,3 +10,13 @@ You should edit this table to contain any new citation tags you create.
 
 [`generated`](generated) contains auto-generated files related to referencing.
 You should not edit these files manually.
+
+[`manual-references.json`](manual-references.json) contains manually managed
+reference data, which overrides the automatically extracted data.
+
+## Reference overrides
```
dhimmel: Wow @agitter you rock at writing docs!

agitter:
> I'm thinking we should also move the citation format info out of CONTRIBUTING to README.

I agree that makes sense.  I'm also opening a new issue with more thoughts on usage.

Issue #1, opened 2017-06-30
Automated figure & table numbering


dhimmel:
@slochower welcome to manubot-rootstock... which is meant to be forked when creating a new manuscript. Still a work in progress.

See previous discussions at https://github.com/greenelab/deep-review/issues/354#issuecomment-306067719 and https://github.com/greenelab/deep-review/issues/558.

It seems like the best way to number and reference tables and figures will be with `pandoc-tablenos` and `pandoc-fignos`, which are both python packages by @tomduck that we can add to the environment:

+ https://github.com/tomduck/pandoc-tablenos
+ https://github.com/tomduck/pandoc-fignos

They can be enabled in the pandoc conversion script with:

```
--filter pandoc-fignos
--filter pandoc-tablenos
```

Since we're also using jinja2 templating, we could do the conversion prior to pandoc if there is a compelling reason.

@slochower do you want to submit the PR? I'm thinking the initial use case we should target is markdown tables and figures embedded via absolute URL (let's save the relative image path case for later).

Also @slowchower, any idea how figure and table captions work?

CC @agitter.

dhimmel:
@slowchower, we should also consider [`pandoc-crossref`](https://github.com/lierdakil/pandoc-crossref) by @lierdakil. `pandoc-crossref` appears to be a more versatile numbering solution, supporting tables, figures, equations and more. The downside may be that it's harder to install in our conda environment, since I couldn't find a conda binary and it's not a python package.

dhimmel:
One additional consideration is that we need LaTeX free solutions, as currently we export to HTML and PDF, but not using LaTeX.

slochower:
@dhimmel Excellent; this looks fun. I'll take a look tomorrow and get back with some more information about this and #2.

slochower:
> It seems like the best way to number and reference tables and figures will be with pandoc-tablenos and pandoc-fignos, which are both python packages by @tomduck that we can add to the environment:

Right. These are mostly easy.

> ...we should also consider pandoc-crossref by @lierdakil

That looks nice. I feel like I stumbled across it at some point, but haven't used it. Maybe I should hold off on the PR while we evaluate this?

>  The downside may be that it's harder to install in our conda environment, since I couldn't find a conda binary and it's not a python package.

Do you mean for CI compilation? I think we can get a VM with GHC / `cabal` available. I don't remember what image the travis build is currently using.

[...]

Issue #2, opened 2017-06-30
Adopting the pandoc-citeproc markdown citation format


dhimmel:
Currently we cite multiple documents like:

```md
Several groups [@doi:10.1371/journal.pone.0032235 @doi:10.1109/TCBB.2014.2343960 @doi:10.1038/srep11476] initiated
```

Prior to pandoc, this gets converted to:

```md
Several groups [@1AlhRKQbe; @ZzaRyGuJ; @UpFrhdJf] initiated
```

Then post pandoc conversion, it will look like:

```
Several groups [30,192,193] initiated
```

Note how we have to add semicolons to separate each reference. We figured this out at https://github.com/lierdakil/pandoc-crossref/issues/110. It would be nice to align our format with the `pandoc-citeproc` format. This presumably would also allow us to make non-bracketed citations like:

```
@doi:10.1371/journal.pone.0032235 was the first group 
```

This would presumably render to 

```
Qi et al 2012 was the first group
```

However, I haven't found the actual docs for the markdown citation formatting supported by `pandoc-citeproc` ([docs](https://github.com/jgm/pandoc-citeproc/blob/master/man/pandoc-citeproc.1.md)). Tagging @lierdakil and @slochower in case they have any insights.

dhimmel:
@jgm (creator of `pandoc-citeproc`): could you point us to where we can find more info on the syntax for markdown citations?

dhimmel:
Just found the following in the pandoc docs. Converted to markdown using this [cool tool](https://euangoddard.github.io/clipboard2markdown/).

***

Citations go inside square brackets and are separated by semicolons. Each citation must have a key, composed of '@' + the citation identifier from the database, and may optionally have a prefix, a locator, and a suffix. The citation key must begin with a letter, digit, or `_`, and may contain alphanumerics, `_`, and internal punctuation characters (`:.#$%&-+?<>~/`). Here are some examples:

```
Blah blah [see @doe99, pp. 33-35; also @smith04, chap. 1].

Blah blah [@doe99, pp. 33-35, 38-39 and *passim*].

Blah blah [@smith04; @doe99].
```

[...]

PR #8, opened 2017-07-07
Automated figure & table numbering (#1)

slochower:
Add the ability to automatically number figures and tables.

Unfortunately, the cross-reference syntax is a little ugly because of `jinja`. In order to prevent `jinja` from interpreting the equation and figure labels, which are inside `{` and `}`, we need to escape the tags. To do this introduces additional complexity, which is mentioned in the manuscript text. Namely, `{#eq:label}` becomes `{{ '{#eq:label}' }}`. There may be a more elegant way around this, but for now, the basic functionality should work as expected. Also, for now I've placed the images in `sections/images` and to build the PDF, I first change to the `output/` directory so relative paths work.


sections/01_abstract.md
```diff
@@ -3,3 +3,14 @@
...
+
+Equations can be integrated and labeled like this `{{ '{{ \'{#eq:label}\' }}' }}` and then
+referenced using `{@eq:label}` like this for Equation @eq:line.
+
+$$ y = mx + b $$ {{ '{#eq:line}' }}
+
+Images can be included and labeled like `{{ '{{ \'{#fig:label}\' }}' }}` and then
+referenced using `{@fig:label}` like this for Figure @fig:lotus.
+
+![An example image.](./images/lotus.jpg){{ '{#fig:lotus}' }}
```
dhimmel: Does `images/lotus.jpg` work instead of `./images/lotus.jpg`? Assuming those are the same?

If I'm understanding correctly, this relative path is just passed through... no modifications?

sections/01_abstract.md
```diff
@@ -3,3 +3,14 @@
 Offspring of the [Deep Review](https://github.com/greenelab/deep-review).
 We cite the Deep Review like [@doi:10.1101/142760] or [@tag:deep_review].
 The source repository is [@url:https://github.com/greenelab/manubot-rootstock].
+
+Equations can be integrated and labeled like this `{{ '{{ \'{#eq:label}\' }}' }}` and then
+referenced using `{@eq:label}` like this for Equation @eq:line.
+
+$$ y = mx + b $$ {{ '{#eq:line}' }}
```
dhimmel: Just to make sure I understand, equations were supported before... this PR is just adding the equation numbering?

sections/01_abstract.md
```diff
@@ -3,3 +3,14 @@
...
+
+Images can be included and labeled like `{{ '{{ \'{#fig:label}\' }}' }}` and then
+referenced using `{@fig:label}` like this for Figure @fig:lotus.
```
dhimmel: Regarding `{@fig:label}`: If [my reading](https://github.com/tomduck/pandoc-fignos#markdown-syntax) is correct, the brackets around image references is not needed. Let's not use them?

sections/01_abstract.md
```diff
@@ -3,3 +3,14 @@
 Offspring of the [Deep Review](https://github.com/greenelab/deep-review).
 We cite the Deep Review like [@doi:10.1101/142760] or [@tag:deep_review].
 The source repository is [@url:https://github.com/greenelab/manubot-rootstock].
+
+Equations can be integrated and labeled like this `{{ '{{ \'{#eq:label}\' }}' }}` and then
+referenced using `{@eq:label}` like this for Equation @eq:line.
+
+$$ y = mx + b $$ {{ '{#eq:line}' }}
+
+Images can be included and labeled like `{{ '{{ \'{#fig:label}\' }}' }}` and then
```
dhimmel: It looks like the bracket syntax is not unique to pandoc-fignos. Instead, it's a markdown extension called `bracketed_spans` and we should respect them. We'll have other uses, for example setting figure dimensions.

I'd prefer jinja2's [raw block](http://jinja.pocoo.org/docs/2.9/templates/#escaping) for readability than this.

dhimmel: Wow a bunch of stuff I wrote here didn't save... bummer. Basically, I think we can find a solution with jinja.

Does `{ #fig:label}` not work... wouldn't that avoid the opening [comment](http://jinja.pocoo.org/docs/2.9/templates/#comments)? Could we override comment_start_string in our [jinja2 environment](http://jinja.pocoo.org/docs/2.9/api/#jinja2.Environment)? For example, `comment_start_string='{##'`?

build/build.sh
```diff
@@ -11,12 +11,16 @@ INPUT_PATH=references/generated/all-sections.md
...
 mkdir -p output
+# Move images to output directory
+cp -R sections/images output/
 
 # Create HTML outpout
 # http://pandoc.org/MANUAL.html
 echo "Exporting HTML manuscript"
 pandoc --verbose \
   --from=markdown --to=html \
+  --filter pandoc-fignos \
```
slochower: Fair point, I don't think this will be too much trouble.

build/build.sh
```diff
@@ -28,11 +32,14 @@ pandoc --verbose \
 
 # Create PDF outpout
 echo "Exporting PDF manuscript"
+cd output
```
slochower: There may be a way around this, but the PDF is already being built in a different location than `sections/` (and `sections/images`) so any path used to reference the images in the Markdown will also have to work for the PDFs. That is, if the Markdown file references `.images/image.png` then the PDF converter must also be able to find that. Any thoughts?

sections/01_abstract.md
slochower: Yup.

sections/01_abstract.md
slochower: It'd be nice if we can find a way with `jinja`. Based on my testing earlier today, adding a space (e.g. `{ #eq:line}`) did not work for equation labels (I didn't test here with a space for figure labels). I saw `comment_start_string` in the documentation but wasn't sure if (or how) we'd like to implement them.

sections/01_abstract.md
slochower: That's what I think too, but I wanted to cover the bases with supporting the more complicated before -- if possible -- so we could claim full compatibility (and so someone familiar with those packages could use this template without switching things up).

sections/01_abstract.md
slochower: I think so, yes.

build/build.sh
```diff
@@ -11,12 +11,16 @@ INPUT_PATH=references/generated/all-sections.md
 
 # Make output directory
 mkdir -p output
+# Move images to output directory
+cp -R sections/images output/
```
slochower: I think that would work probably. Would the user have to commit the images to the output branch and the main text to the master branch?

slochower:
> It'd be nice to also add am image by URL. Image by versioned github URL should be the preferred method on including images... since it shows image provenance. For most science studies, there will be other repos that create the figures. 

Fair point.
> For example here, were is this lotus figure from. Is it CC0 licensed / public domain? If not, we need another image or to comply with image license. In general, image source should go in the commit message that adds it.

https://en.wikipedia.org/wiki/Rhizome#/media/File:Lotus_root.jpg (CC BY-SA 2.0)
It seems there are some CC0 images here: https://www.pexels.com/

dhimmel:
Not ideal to use a CC BY-SA image in a CC BY work... at a minimum you must explicitly state the source and mention it's CC BY-SA. Better if image is CC BY or ideally CC0 (no worries).

If you want you can use [any of the images](https://github.com/greenelab/scihub/tree/master/figure) from our forthcoming Sci-Hub study... all CC0. Here's markdown for Figure 1:

```md
![Google Trends Search interest for Sci-Hub and LibGen](https://cdn.rawgit.com/greenelab/scihub/7891082161dbcfcd5eeb1d7b76ee99ab44b95064/explore/trends/google-trends.svg)
```

Also anything from pexels or pixabay would work.

build/build.sh
dhimmel: Okay after some research, this makes me think we should be building from the `output` directory if we're going to be using relative links. If we move track `output/images` and put images there, then relative links should work for PDF and HTML exports. Downside is that figures won't show up in text editor. Your original copy directory method would solve this, but I don't like the duplication... the repo structure is already complicated.

What do you think about only supporting images by URL for now? One downside is you need a network connection... but you already need that to build. Also harder to have images hosted in private repos. I'm on the fence here... @agitter and @slochower what are your thoughts?

build/citations.py
```diff
@@ -50,9 +50,10 @@ def semicolon_separate_references(text):
 def get_brackets_without_reference(text):
     """
     Find bracketed text that does not start with @. Does not match
-    brackets that are followed by an open parenthesis.
+    brackets that are followed by an open parenthesis. Also ignore
+    brackets that start with ! for captions.
     """
-    pattern = re.compile(r'(\[[^@][^\]]*?\])[^(]', flags=re.DOTALL)
+    pattern = re.compile(r'(?<!\!)(\[(?!\@).+\])[^(]', flags=re.DOTALL)
```
dhimmel: This part will hopefully be superseded if we address #2 and no longer have to worry about brackets at all?

sections/01_abstract.md
dhimmel: > wasn't sure if (or how) we'd like to implement them

I think definitely yes. I'm assuming you'd specify the jinja2 [environment here](https://github.com/greenelab/manubot-rootstock/blob/439bf03bd49e0ccf39c7e1689a71daa445c21400/build/references.py#L171-L173). Let me know if you get stuck and I can look more into it.

sections/01_abstract.md
dhimmel: > That's what I think too, but I wanted to cover the bases with supporting the more complicated before

Let's go with the simple method, as I think its universally preferable and makes fewer brackets which appear to be pain points. If we really want to make sure we cover the corner cases, we should implement test manuscripts, rather than put the non-preferred usage in the example.

agitter:
I'm not concerned about the network connectivity requirement for building.
Making it harder to build with private images in private repositories could
eventually be a concern.  But until someone demonstrates that need or
requests that feature, I'm okay with URL only images.

It's minor, but we may want to move these new examples out of the abstract
and into a new .md file.

build/build.sh
dhimmel: > Would the user have to commit the images to the output branch and the main text to the master branch?

`output/images` would be a tracked directory on the master branch. It would also automatically get exported to the `gh-pages` branch via ghp-import.

slochower:
I'm okay with supporting URL images but I'm not sure what it would mean to _not support_ file based images, unless that just means not creating an `images/` subdirectory in the default repository. AFAIK, images via URLs should just work (although I have not tested it yet).

dhimmel:
> we may want to move these new examples out of the abstract and into a new .md file.

Good point. Perhaps something like: `02_examples.md`. @slochower your call.

> I'm okay with supporting URL images but I'm not sure what it would mean to not support file based images, unless that just means not creating an images/ subdirectory in the default repository.

Okay. Let's proceed without path-based images for now. This is something we can revisit in the future should there be demand. Note that you can always have the images in the repo, and just refer to them via GitHub URL. I do think we have probably have figured out the optimal way to include path-based images... but that the complexity may not be worth it, as well as the advantages to properly provenanced images that reside in the repos that generate them.

slochower:
@dhimmel agreed & I'll move the examples to another file. Probably will work on this sporadically in the following few days.

dhimmel:
> I'll move the examples to another file. Probably will work on this sporadically in the following few days

Sounds good! Thanks for dealing with the barrage of review comments.

agitter:
> I'm not sure what it would mean to not support file based images

@slochower I think not supporting filed-based images would simply mean not documenting that functionality.  Eventually we'll need to document all of the features available in this manuscript build system that are not standard GitHub Flavored Markdown.

@dhimmel What happens if we have an image added by URL that is no longer available?  Will we have a cached version of it in the gh-pages branch that we could switch to, still giving attribution to the original source in text?

dhimmel:
> Eventually we'll need to document all of the features available in this manuscript build system that are not standard GitHub Flavored Markdown.

Great point!

> What happens if we have an image added by URL that is no longer available?

The build would either fail or have a missing image. The PDF may have a cached version, but the HTML version probably just always loads from the URL.

For this reason, it's important for longevity to only use URLs that will be persistent. For example, images should primarily be hosted in GitHub repos controlled by the manuscript authors. Therefore, you minimize and align the failure modes... If you want to use an image from elsewhere, I'd recommend tracking it in a repo that you control.

build/references.py
```diff
@@ -169,7 +169,10 @@ def get_standard_citatation(citation, cache, override):
 converted_text = semicolon_separate_references(converted_text)
 
 # Template using jina2
-template = jinja2.Template(converted_text)
+#env = jinja2.Environment(comment_start_string='{##')
+#template = env.get_template(converted_text)
+template = jinja2.Environment(loader=jinja2.BaseLoader(),
+                              comment_start_string='{##').from_string(converted_text)
```
dhimmel: For readability, I think it's best to have `.from_string(converted_text)` or `template = env.get_template(converted_text)` on its own line.

Also remove commented out code.

sections/04_tables.md
```diff
@@ -0,0 +1,15 @@
+# Tables
+
+Numbered tables can be included by using Markdown syntax to create the
+table and then adding a tag after the caption like this `{#tbl:label}`. The
+tables can be referenced in the text by using `@tbl:label`.
+
+Table @tbl:interest shows the relative search interest of the terms "Sci-Hub"
+and "LibGen" the week of June 25, 2017.
+
+week       search_term interest
```
dhimmel: Let's use a markdown formatted table as generated by http://www.tablesgenerator.com/markdown_tables. I think this is more standard and robust than the format here

sections/03_figures.md
```diff
@@ -0,0 +1,10 @@
+## Figures
+
+Numbered figures can be included by using the format
```
dhimmel: Do one sentence per line rather than wrap at character limit.

dhimmel:
@slochower if you build this locally can you drag the PDF into a comment here. Would like to see how that looks.

slochower:
I added commit add91f1 to address your comments tonight. HTML rendering looks great (to me) but the PDF is not as nice. The most obvious mistake is that the image, now rendered via URL, is missing. Some of the spacing also hasn't transferred well from HTML.
[manuscript.pdf](https://github.com/greenelab/manubot-rootstock/files/1140822/manuscript.pdf)

slochower:
This [thread](https://github.com/jgm/pandoc/issues/750) possibly related to the missing images issues, although I haven't thoroughly investigated.

slochower:
> Remove lotus.jpg

@dhimmel I did a `git rm` on it, but it's still showing on this PR. It's unclear to me if that's the expected behavior.

build/references.py
```diff
@@ -169,7 +169,9 @@ def get_standard_citatation(citation, cache, override):
 converted_text = semicolon_separate_references(converted_text)
 
 # Template using jina2
-template = jinja2.Template(converted_text)
+env = jinja2.Environment(loader=jinja2.BaseLoader(),
```
dhimmel: Will this all fit on one line:

```
env = jinja2.Environment(loader=jinja2.BaseLoader(), comment_start_string='{##')
```

Or align to make flake8 happy.

slochower: Oh, that was ugly before. It is exactly 80 characters, so yes.

slochower:
Yes, I did a commit and push. Is there a way to use a style sheet for the PDF generation? I think things like `code` text should be handled differently, for example. This is probably a task for another PR, though.

slochower:
Even spacing between the text and the references is different between the PDF and HTML for reasons that are unclear to me. It sure looks like the PDF is missing a space between the text and reference even though the HTML is fine and they are built from the same source.

PDF
image
HTML
image

dhimmel:
@slochower this is ready to merge besides the lotus.jpg. Still having trouble with the deletion?

slochower:
@dhimmel yes, actually. Yesterday, before 026cf87 I did a local `git rm` on lotus.jpg. I don't even have a local copy anymore (i.e., I did not use `--cached`) and my working tree is clean.

dhimmel:
Okay I'll merge this PR manually and have a look.

dhimmel:
I merged this manually in https://github.com/greenelab/manubot-rootstock/commit/b03e1c33efb616b696fda22662af240a9cc061c9. @slochower, I'd recommend making your master branch always follow the upstream (`greenelab:master`) and creating new branches for pull requests. Before making anymore changes, make sure to start from the latest commit in `greenelab:master`.

Also check that everything built correctly after the merge (still building ATM). Thanks for the contribution!

dhimmel:
> Even spacing between the text and the references is different between the PDF and HTML for reasons that are unclear to me. It sure looks like the PDF is missing a space between the text and reference even though the HTML is fine and they are built from the same source.

Feel free to open an issue along the lines of "why does the PDF formatting suck so much?"

Issue #6, opened 2017-07-05
Forking a new manuscript documentation


agitter:
I started thinking about more detailed documentation for someone who wanted to create a new manuscript using this repository as a template.  They could fork through GitHub, but that would only support a single manuscript per user.

The process I'm trying is roughly:
- Clone https://github.com/greenelab/manubot-rootstock locally and rename `manubot-rootstock` to my desired new manuscript name
- Create a new repository on GitHub with that name
- Set my new GitHub remote as `origin` and `manubot-rootstock` as `upstream`
- Enable Travis CI and change the readme links
- Change the github.io links
- Execute some (or all?) of `initialize.sh` to create the necessary remote branches
- Generate and configure the GitHub SSH deploy key for `deploy.sh`

Is there are more streamlined process we could recommend?  Am I missing any steps?

dhimmel:
As you identify, there are still quite a few steps to initializing this thing.

Some of the steps we can automate... like SSH key generation and replacing hyperlinks. We can probably make `initialize.sh` do these things.

> Set my new GitHub remote as `origin` and `manubot-rootstock` as `upstream`

I'd probably set `greenelab:manubot-rootstock` as `rootstock`. Remember, upstream will refer to your clone, when you're working off a fork. I'm still not sure how easy it will be to merge in new changes to `greenelab:manubot-rootstock`. I guess we'll see. 

> They could fork through GitHub, but that would only support a single manuscript per user.

I didn't realize that there's a one fork per user limitation. Too bad... cloning it will be.

I'm going to create a clone for our Sci-Hub manuscript soon from https://github.com/greenelab/scihub/pull/18. Therefore, I should be able to clarify and improve many of this workflow.

agitter:
Extending `initialize.sh` in the future would certainly make parts of this more convenient.  The biggest bottleneck for me so far was installing the Travis CI Command Line Client.  It requires Ruby 1.9.3 or above, which is newer than the most recent version available on some of my older Linux systems' package managers.  I haven't had time to look for a workaround.

agitter:
If anyone else has the same problem I did with the Travis CI client, running it in Docker may be the easiest solution: https://github.com/andredumas/docker-travis-ci-cli

agitter:
My [toy example of a cloned manuscript](https://github.com/agitter/my-manuscript) works now.  After we extend the initialization script, I can help document the cloning process.

dhimmel:
> My toy example of a cloned manuscript works now

[...]

PR #9, opened 2017-07-10
Update shortened URLs

agitter:
While working through the process outlined in #6 I noticed the short URLs still pointed to deep-review


dhimmel:
Eventually, the initialization script should automatically [generate](https://github.com/blog/985-git-io-github-url-shortener) these short URLs.

PR #10, opened 2017-07-10
Rename sections to content and restore author inclusion

dhimmel:
Does not fill out pandoc author metadata. Leaving that for @evancofer as per https://github.com/greenelab/manubot-rootstock/issues/4#issuecomment-313883585.

The PDF now gets the following:

pdf-authors

The ugliness, large text, and large margins in the PDF is a separate issue.


dhimmel:
FYI, https://github.com/greenelab/manubot-rootstock/pull/10/commits/095c931fe4f4fe7fefdc0a5a424e58e46b63c289 is unrelated to authorship parsing.

dhimmel:
@agitter does this look good to you?

build/authors.py
```diff
@@ -2,143 +2,21 @@
...
-    author_df = pandas.read_table(author_file)
-
-    # Add author initials
-    author_df['initials'] = author_df['full_name'].apply(initials)
-
-    # Avoid initial collisions, will need a more general way to refer to
-    # authors if they occur
-    assert len(author_df['initials'].unique()) == len(author_df), \
-        'All authors must have unique initials'
-    assert np.all(author_df['approve'].str.match('Yes', case=False)), \
```
agitter: In the readme we should document that adding oneself to `authors.tsv` implies that the author has read and approved the manuscript.

build/citations.py
```diff
@@ -61,13 +61,10 @@ def get_text(directory):
     Return a dictionary of section texts in the specified directory.
     """
     section_dir = pathlib.Path(directory)
-    paths = sorted(section_dir.iterdir())
+    paths = sorted(section_dir.glob('[0-9]*.md'))
```
agitter: The (existing) requirement that manuscript sections begin with numbers makes sense to me.  This is something else to document in case someone expects all `.md` files to be included.

content/authors.tsv
```diff
@@ -0,0 +1,3 @@
+github_username	full_name	initials	orcid	email	affiliations	funders
```
agitter: Do you still want a symbols column for superscripts?

content/00.front-matter.md
```diff
@@ -0,0 +1,12 @@
+# Manubot Rootstock: Manuscript Title
+
+## Authors
+
+{% for author in authors %}
```
agitter: This is nice and clean.

content/00.front-matter.md
```diff
@@ -0,0 +1,12 @@
...
++ **{{author.full_name}}**<br>
```
agitter: It looks like `<br>` works fine in the pdf based on your screenshot.  Given our desire to primarily support pandoc conversion to html and pdf, this should be fine.  I'm not sure whether it would create problems with docx or latex conversions.

content/00.front-matter.md
```diff
@@ -0,0 +1,12 @@
...
+  [\@{{author.github_username}}](https://github.com/{{author.github_username}}) ·
+  [{{author.orcid}}](https://orcid.org/{{author.orcid}}) ·
+  {{author.affiliations}} ·
+  Funded by {{author.funders}}
```
agitter: In a full manuscript, this would be better suited for a dedicated funding section in the "back matter".  For this example manuscript, I think it is fine to leave here.  Users should be able to figure out how to loop over the authors again if they want to move the funding information.

content/authors.tsv
dhimmel: I was planning to leave that for a future PR, when the needs were more clear.

content/00.front-matter.md
dhimmel: One possibility would be to put this author section at the end of the manuscript and then just include an author list in the beginning. Leaving this decision to @evencofer if he submits a PR to address the author metadata as per https://github.com/greenelab/manubot-rootstock/issues/4#issuecomment-313883585.

agitter: That might work well.  We can merge this and defer that decision (cc @evancofer).

build/authors.py
dhimmel: Thanks for enforcing docs upkeep!

agitter: I'm mostly trying to make some notes as we proceed.  We may not want to actually write all the docs until the build system stabilizes.

README.md
```diff
@@ -1,12 +1,22 @@
...
 A rootstock (or rhizome) is "a plant onto which another variety is grafted".
 This repository plays the role of a botanical trunk.
 Fork it and a manuscript will grow.
 
 The most current version of the `master` branch is built by continuous integration and [available via gh-pages](https://greenelab.github.io/manubot-rootstock/).
 To see what's incoming, check the [open pull requests](https://github.com/greenelab/manubot-rootstock/pulls).
 
+## Source
+
+The manuscript source is in located in [`content`](content).
```
agitter: Great edits!  This is very helpful.

In this line, `is in located` -> `is located`

PR #11, opened 2017-07-10
Use same locale for all builds

dhimmel:
Fixes issue reported in https://github.com/greenelab/deep-review/pull/569 where some OSes use a non-UTF-8 default encoding.

CC @yfpeng

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -1,5 +1,8 @@
 set -o errexit
 
+# Default Python to read/write text files using UTF-8 encoding
+export LC_ALL=en_US.UTF-8
+
 # Generate reference information
 echo "Retrieving and processing reference metadata"
 (cd build && python references.py)


build/build.sh
```diff
@@ -1,5 +1,8 @@
 set -o errexit
 
+# Default Python to read/write text files using UTF-8 encoding
+export LC_ALL=en_US.UTF-8
```
yfpeng: Neat solution!

agitter:
In the interest of testing how to keep cloned manuscripts synchronized with the original rootstock, I confirmed I was able to [rebase this change into my repository](https://github.com/agitter/my-manuscript/commit/d52345c9dee8f5b4a9dc31e2af4231aace2d9e5d).  It would get more complicated for pull requests that touch multiple files if only some of them should be merged into the cloned repository.

dhimmel:
> It would get more complicated for pull requests that touch multiple files if only some of them should be merged into the cloned repository.

Yep, we should think about distinguishing files that should be changed from files that shouldn't be changed. And then perhaps, we can release the git commands to only fast-forward the files that shouldn't be changed in a clone.

Issue #4, opened 2017-06-30
Journal compatibility


agitter:
I'm excited to see this standalone manuscript repository!

I have a general question in regards to journal submissions.  Many journals require Word or LaTex formats for submission.  Have you thought about how manuscripts written in this markdown format can be submitted to a journal with those requirements?  Would one use pandoc outside of the automatic build to do a one time conversion to Word or LaTeX?

dhimmel:
You should be able to adapt our pandoc commands to convert to Word or LaTeX. There's a chance some things will break... Perhaps try `--to=docx` or `--to=latex` in the deep review pandoc command.

One other option is to copy the rich text from the rendered HTML and paste that into a word processor. This works surprisingly well, despite being a sort of hack.

slochower:
This is a good point and one I've run into personally a few times. You can definitely get a Word `.docx` out of `pandoc`. Problems crop up when a journal insists the authors to use a Word template or in-house styles (e.g., Science). In my experience, `pandoc` has partial functionality for many cases, but it gets ugly and I'm not sure there is any way around it besides manual tinkering. 

An alternative approach by the `R` folks has been to specifically craft R Markdown templates for specific journals: https://github.com/rstudio/rticles, but I think that approach deviates from the Swiss army knife features of `pandoc`, is a lot of effort, and will become out of date when the publishers change something.

evancofer:
With the current approach, the LaTeX output from Pandoc includes the authors and their affiliations as text in the document body rather than using the built-in "author" field for this. The same issue applies to the title, which is instead placed in as a section. Subsequent sections are placed in as subsections. Lastly, the superscript fields for author affiliations (i.e., ``<sup>``) do not translate from html5 to LaTeX. If this is something you have an interest in addressing, I can contribute a PR for it.

dhimmel:
@evancofer one of the main motivations of to be a next generation manuscript creation system that frees itself from the limitations of LaTeX. So I'm hesitant to start modifying the core implementation for LaTeX compatibility. On the other hand it's nice that to be able to convert from the markdown source to any format that has utility for you...

What's the specific problem you're trying to solve? If it's submission to journal, is the docx export (or copy and RTF paste) not an adequate solution? Going forward I see more journals accepting PDF initial submissions and then markdown or perhaps JATS XML for final submission.

> If this is something you have an interest in addressing, I can contribute a PR for it.

[...]

PR #16, opened 2017-07-12
Added support for yaml_metadata_block for #4

evancofer:
Also some minor spelling fixes, and reassigned sectioning.


evancofer:
Discussion:
- I could not get ``pandoc_title_block`` to embed author metadata, so I stuck to the ``yaml_metadata_block``
- I used [pyyaml](http://pyyaml.org/wiki/PyYAMLDocumentation), as it seemed to be the most pythonic YAML library. Indeed, this method adds the author metadata to the PDF output, but only when LaTeX is used to generate said output. Additionally, this method adds author metadata to the HTML output. This method also preserves keyword metadata in the PDF and HTML output. My only reservation about the HTML output is that it doesn't use the citation tags that I've commonly seen elsewhere (i.e. [these](http://blog.reallywow.com/archives/123/)). I'm sure this could be solved with a simple template.

- The formatting on the HTML output looks odd at this stage, but that is largely because we haven't added styling to the css file for the ``author`` class used by pandoc.

- As far as the affiliations (which aren't showing up right now, [more on this here](https://github.com/jgm/pandoc-templates/pull/235)), we would have to modify the normal pandoc template to add the author affiliations. This should be trivial however. If we just combine the front matter that we already have with such a template, I think that would do the trick, and preserve the author metadata being created. 

I've attached copies of the PDF and HTML outputs here:
- [manuscript.pdf](https://github.com/greenelab/manubot-rootstock/files/1140737/manuscript.pdf)
- [manuscript-latex.pdf](https://github.com/greenelab/manubot-rootstock/files/1140738/manuscript-latex.pdf)
- [index.txt](https://github.com/greenelab/manubot-rootstock/files/1140740/index.txt) (*I had to change the extension to .txt to upload it on github*)

build/references.py
```diff
@@ -171,8 +172,33 @@ def get_standard_citatation(citation, cache, override):
 template = jinja2.Template(converted_text)
 converted_text = template.render(**stats)
 
+all_sections_file = gen_dir.joinpath('all-sections.md')
+
+#FIXME: Stubbing metadata:
+title_text = "Manubot Rootstock: Manuscript Title"
```
dhimmel: No metadata should be hardcoded in `references.py`. Modify `references.py` sparingly. Instead could you create a `content/metadata.yml` file, with the hardcoded data. Then author information could be inserted into the yaml structure before it's put into `all_sections.md`?

evancofer: I realize that no metadata should be hardcoded. I added a [stub](https://en.wikipedia.org/wiki/Test_stub) to avoid significantly altering project layout without some direction. Where in the build process should I add the call to metadata.py (the file where I am going to move this to)? Should metadata be conditionally supported in the build process?

evancofer:
``yaml_metadata_block`` is being populated for both PDF and HTML output. For some reason, non-LaTeX PDF output is not including it in the document metadata, but everything else is including it in document metadata. I assume it will require some additional work to figure out what is happening there. 
Next step will be to add a new python script (``metadata.py``) to the build process, as well as the ``metadata.yaml`` file to contain the relevant metadata information. I will drop the tex step from this as well. Will work on this later today.

evancofer:
@dhimmel New commit. I've put all the author data in a single metadata.yaml file in content. ``authors.py`` is now a wrapper that accesses this metadata file; I have intentionally done nothing to redesign how author info that was in ``authors.tsv`` is otherwised accessed. Authors & keywords are showing up in HTML output header. I've also gone ahead and removed TeX support. See attached files for outputs:
 - [index.txt](https://github.com/greenelab/manubot-rootstock/files/1143769/index.txt)
 - [manuscript.pdf](https://github.com/greenelab/manubot-rootstock/files/1143768/manuscript.pdf)

agitter:
From a usability perspective, are authors going to have to manually ensure that the title and abstract in `metadata.yml` remain consistent with the title and abstract in `00.front-matter.md` and `01.abstract.md`?  That isn't so terrible.  Authors currently have to manually re-enter that information into forms during journal submission.

An alternative would be to populate `00.front-matter.md` and `01.abstract.md` with the metadata from `metadata.yml`, but I find that to be less user-friendly.  I'm willing to be convinced otherwise.

evancofer:
I think that populating front matter and abstract with the information from metadata.yml is feasible. An alternative is to remove the abstract from ``metadata.yml`` entirely. Apart from the previous LaTeX output, it seems like the abstract metadata is not being used to generate the manuscript output, and I couldn't find evidence of it being used in the HTML metadata header. Including the abstract in the metadata was somewhat unwieldy, but that seems more appropriate for the other information (e.g., author and affiliation). Thus, we could consider dropping ``00.front-matter.md``, but keeping ``01-abstract.md``.

agitter:
If the abstract in `metadata.yml` is not being used outside LaTeX, then removing it seems like a good idea.  That would need to be updated more than the title.

build/metadata.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def get_metadata_object(path):
+    """
+    Load all of the metadata from a yaml file, & return.
+
+    :param path: pathlib.Path to a metadata YAML file.
+    :return: structured metadata object.
+    """
+    class metadata(object):
```
dhimmel: What is the point of a class here? Are you sure that a `dict` or `collections.OrderedDict` is not more appropriate?

content/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+        funders: GBMF4552
+    -
+        github_username: agitter
+        full_name: Anthony Gitter
+        initials: AG
+        orcid: 0000-0002-5324-9833
+        email: gitter@biostat.wisc.edu
+        affiliations: Department of Biostatistics and Medical Informatics, University of Wisconsin-Madison and Morgridge Institute for Research
+        funders: NIH U54AI117924
+keywords: ['work-in-progress', 'markdown', 'manuscript', 'publishing']
```
dhimmel: can each keyword go on a newline for improved diffing?

evancofer: Yes. Like such:
```yaml
keywords:
    - 'work-in-progress'
    - 'markdown'
    - 'manuscript'
    - 'publishing'
```

dhimmel:
@agitter what do you think about replacing `authors.tsv` with a `author_info` object in `metadata.yaml`? Note that pandoc needs an `authors` field with listed authors in `metadata.yaml`, so we will still need to do some processing. The question is whether yaml or TSV is a preferred format for entering author information. YAML has less enforcement... with TSV we specify the columns, but is a bit more human readable.

build/metadata.py
evancofer: I am sure. The object being returned here is not meant to have key-value pairs added to it, so a dictionary type is overkill. While a dictionary type might provide similar functionality, it also means that membership is typically determined at run time, and since python doesn't have static or const, widely used entry-level static code analysis tools (e.g., pylint) won't catch the erroneous call in the following scenario:
```python
metadata['header_block']
metadata['header_blck'] # error at run time
```
Whereas in the following scenario they do:
```python
metadata.header_block
metadata.header_blck # easily catchable before run time
```
Since there is no reason to extend the contents of the metadata dict beyond what would be found in the file itself (I imagine that if you wanted to have empty optional fields you would want them represented by ```None```), and since not using a dict enhances ability to debug code, I decided to not use a dict.

content/metadata.yaml
evancofer: I believe that it will also still work if leave out the quotation marks, which is nice.

agitter:
@dhimmel I think yaml is fine for inputting author information and perhaps even preferable to tsv.  If the title and keywords going to be in `metadata.yml`, then it make sense to put as much metadata there as possible instead of splitting it across two files.  Arguably it is easier for an author to add themselves via a GitHub edit with yaml instead of tsv.  They don't need to track the column headers and get the spacing right if their entry is missing values for some columns.  We can also leave comments in the yaml file showing what author_info fields are available and enforce the require fields as in #17.

It would be easier to re-order long author lists in tsv though.  And someone doing bulk entry of many authors might prefer tsv.

vsmalladi:
@agitter @dhimmel what if we used a json object for the authors?

agitter:
@vsmalladi Are you thinking of using the CSL JSON format?  It would have the advantage of reusing a standard file format.  However, I think it would be cumbersome to update manually.

vsmalladi:
@agitter That is what I was thinking. But I can see how adding more authors would be cumbersome.

dhimmel:
> what if we used a json object for the authors?

I prefer YAML over JSON because there's much less boilerplate. It's worth noting that "JSON is YAML, but YAML is not JSON".

@evancofer go ahead and remove `authors.tsv` and encode this information in `metadata.yaml`.

build/metadata.py
dhimmel: This class has to go. It's not pythonic and confusing.

If you really want immutability (which I don't think you do), you can look into a [`namedtuple`](https://docs.python.org/3/library/collections.html#collections.namedtuple).

evancofer: Yeah, the immutability was not really the point there. I'll make the changes this afternoon.

evancofer:
@dhimmel Changes have been made.

build/references.py
```diff
@@ -171,8 +173,15 @@ def get_standard_citatation(citation, cache, override):
 template = jinja2.Template(converted_text)
 converted_text = template.render(**stats)
 
+# Fetch metadata
+metadata_dict = get_metadata_info(pathlib.Path('../content/metadata.yaml'))
```
dhimmel: Call this `pandoc_metadata` rather than `metadata_dict` to be more explicit.

build/metadata.py
```diff
@@ -0,0 +1,45 @@
...
+        initials.append(author['initials'])
+        metadata_dict['author'].append(author['full_name'])
+    initials_counts = collections.Counter(initials)
+    if max(initials_counts.values()) > 1:
+        duplicated_initials = [k for k, v in d.items() if v > 1]
+        msg = f'Duplicated initials in metadata.yaml: {duplicated_initials}'
+        raise ValueError(msg)
+    metadata = dict()
+    metadata['author_info'] = metadata_dict.pop('author_info')
+    metadata['header_block'] = metadata_dict
```
dhimmel: What is `header_block`?

build/metadata.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python
```
dhimmel: Bet rid of this shebang. Not a script.

build/metadata.py
evancofer: the subset of the metadata that is currently used in yaml_metadata_block

build/metadata.py
evancofer: It's getting moved into references.py so this is totally redundant.

build/metadata.py
```diff
@@ -0,0 +1,45 @@
...
+    if max(initials_counts.values()) > 1:
+        duplicated_initials = [k for k, v in d.items() if v > 1]
+        msg = f'Duplicated initials in metadata.yaml: {duplicated_initials}'
+        raise ValueError(msg)
+    metadata = dict()
+    metadata['author_info'] = metadata_dict.pop('author_info')
+    metadata['header_block'] = metadata_dict
+    return metadata
+
+def get_metadata_info(path):
```
evancofer: That was literally my original implementation ... but you asked me to change it ...

evancofer: I agree 100%. Original implementation was a function inside of ``references.py`` that did exactly what you have described. The current implementation is the way it is to preserve the ``get_author_info`` interface to author information and minimize changes in ``references.py`` ...

build/references.py
```diff
@@ -171,8 +173,15 @@ def get_standard_citatation(citation, cache, override):
...
+
 # Write manuscript for pandoc
-gen_dir.joinpath('all-sections.md').write_text(converted_text)
+all_sections_file = gen_dir.joinpath('all-sections.md')
```
evancofer: Can't a path refer to a directory and a file?

build/metadata.py
evancofer: If that's the case, we should git rid of both authors.py and metadata.py.

build/metadata.py
evancofer: Not all of the metadata (e.g., funding) is of any use for the ``yaml_metadata_block``. Similarly, not all of the yaml metadata is used by the funding statements, etc. Therefore, I keep them distinct.

build/references.py
evancofer: Good catch.

build/metadata.py
dhimmel: `metadata.py` can go. `authors.py` could go, but we don't want to add too much complexity to `references.py`. Therefore, if you want to perform validation of author data, perhaps keep `authors.py` and create a function there for `validate_authors`.

build/metadata.py
dhimmel: I don't think there is a big downside to passing the entire `metadata.yaml` with `authors` added to `yaml_metadata_block`. Just thinking of ways to minimize the code complexity

build/metadata.py
evancofer: I'm could just leave ``metadata.py``, and have ``validation_authors`` in it. I would also add ``validate_pandoc_metadata`` to it. It can hold our general methods for validating different types of metadata, etc. Is that okay?

build/metadata.py
evancofer: The problem is that the way authors dict has been handled in code elsewhere might not work with the ``yaml_metadata_block`` formatting. The real issue is that ``yaml_metadata_block`` wants ``dict['full_name']`` to instead be called ``dict['author']``, whereas that's not our current approach anywhere else.

dhimmel: In any case, the list of full names to go in `pandoc_metadata['author']` will be created in Python from  `metadata['author_info']['full_name']`. I don't think it will cause an issue to pass extra metadata to pandoc... especially since oftentimes that will be a feature. The question is do we know now all the metadata fields we want to passthrough? I don't think we do, so it may be best to passthrough everything and just make sure not to define fields that are used by pandoc unless that is the intention. For example, pandoc-tablenos has metadata options like `xnos-section-numbers: On`. This we would want to passthrough.

> Lastly, we may want to decouple the validation of the metadata (e.g., title) and the validation of the author information (e.g., names) to allow for double blind peer-review if necessary.

Double blind peer review is not supported by this system.

build/metadata.py
dhimmel: Let's save more general validation for a future PR. I'd like to get this PR merged soon. I'm even fine with skipping all metadata validation at this point.

build/build.sh
```diff
@@ -32,10 +34,10 @@ pandoc --verbose \
   --output=output/index.html \
   $INPUT_PATH
 
-# Create PDF outpout
+# Create PDF output
 echo "Exporting PDF manuscript"
 pandoc \
-  --from=markdown \
+  --from=markdown+yaml_metadata_block \
```
dhimmel: As per http://pandoc.org/MANUAL.html, I believe the `yaml_metadata_block` is on by default.

build/build.sh
```diff
@@ -5,21 +5,23 @@ export LC_ALL=en_US.UTF-8
 
 # Generate reference information
 echo "Retrieving and processing reference metadata"
-(cd build && python references.py)
+(cd build && python references.py && python metadata.py)
```
dhimmel: `&& python metadata.py` should not be needed?

dhimmel: Ah looks like you already addressed this.

build/build.sh
evancofer: So it is.

build/environment.yml
```diff
@@ -20,3 +21,6 @@ dependencies:
     - pybase62==0.3.2
     - pysha3==1.0.2
     - python-bitcoinlib==0.7.0
+    - pandoc-fignos==0.20
```
dhimmel: Remove these additions (already in this list)

content/00.front-matter.md
```diff
@@ -1,6 +1,5 @@
-# Manubot Rootstock: Manuscript Title
 
-## Authors
+# Authors
```
dhimmel: In the resulting HTML, what heading level do these render to. If h1 and title is also h1, don't we want to keep these as level 2?

evancofer: Yeah. It will actually work better with them at level 2 under the title. It will allow you to have supplementary content as a level 1 header at end, with level 2's under it. That way, in the pdf navigation, the top levels are broken down into the major divisions of the manuscript.

evancofer:
@dhimmel yeah, it's been merged. I also changed the approach to checking author formatting (e.g., no duplicate initials) to work with the YAML-based metadata dict.

content/00.front-matter.md
```diff
@@ -1,4 +1,3 @@
-# Manubot Rootstock: Manuscript Title
```
agitter: How is the title added now?  I expected we would need a template here.

evancofer: The title is being added through the metadata. This is part of the default functionality of the ``yaml_metadata_block`` markdown extension.

agitter: Very cool!  Thanks.

Issue #15, opened 2017-07-12
nan (missing) author fields


dhimmel:
Quoting from https://greenelab.github.io/scihub-manuscript/

> 0000-0002-9925-9623 · Department of Applied Bioinformatics, Institute of Cell Biology and Neuroscience, Goethe University Frankfurt · Funded by nan

Would be better to have jinja2 omit blank fields entirely. In other words remove " · Funded by nan"

vsmalladi:
Should be able to update 00.front-matter.md to be conditional on fields.

dhimmel:
> Should be able to update 00.front-matter.md to be conditional on fields.

Yep. want to submit a PR?

vsmalladi:
Ya I can do that. Just testing the code locally.

vsmalladi:
Are there any fields you want required?
 I was thinking:
 

- github_username
- full_name	
- initials	
- orcid
- email

dhimmel:
> Are there any fields you want required?

What does required mean? Enforced in the source code or just always displayed even if blank?

As far as truly required, probably name and initials only. But I'm mostly concerned about fixing the output, rather than automating checks that PR review should catch. Different projects will have different needs when it comes to this metadata.

vsmalladi:
Okay that makes sense. I will just have Name and inititals present then otherwise if blank don't display.

slochower:
This PR seems to create a references warning for me. I believe the the code here works and the code to detect a malformed reference should be adjusted, but I admit I haven't had time to check yet. I'm also not sure why this didn't pop up on Travis CI here, but did for my fork (perhaps because of the different entries in `authors.tsv`).

![image](https://user-images.githubusercontent.com/5393995/28236002-a78b2b30-68cf-11e7-98d9-01df6f02bff9.png)

dhimmel:
Yeah, I noticed this too. The builds come out fine... we should relax the malformed reference detection. Or entirely remove it if we pursue #2.

PR #17, opened 2017-07-12
Fix formatting of missing author fields #15

vsmalladi:
Put checks in for Name and initials to be required. Add in conditional… formatting if fields are missing.

### DIFF
--- a/build/authors.py
+++ b/build/authors.py
@@ -31,7 +31,7 @@ def get_author_info(path):
         raise ValueError(msg)
 
     # Replace missing values with None
-    #author_df.fillna("None", inplace=True)
+    author_df.where((pandas.notnull(author_df)), None,inplace=True)
 
     authors = author_df.to_dict(orient='records')
     return authors
--- a/build/authors.py
+++ b/build/authors.py
@@ -12,15 +12,15 @@ def get_author_info(path):
     author_df = pandas.read_table(path)
 
     # Detect missing full name
-    missing_full_name = list(numpy.where(author_df['full_name'].isnull())[0])
+    missing_full_name = author_df[author_df['full_name'].isnull()].index.tolist()
     if missing_full_name:
         msg = f'Missing full name in authors.tsv for author(s): {missing_full_name}'
         raise ValueError(msg)
 
     # Detect missing author initials
-    missing_initials = list(numpy.where(author_df['initials'].isnull())[0])
+    missing_initials = author_df[author_df['initials'].isnull()].index.tolist()
     if missing_initials:
-        msg = f'Missing initials in authors.tsv for author(s): {missing_full_name}'
+        msg = f'Missing initials in authors.tsv for author(s): {missing_initials}'
         raise ValueError(msg)
 
     # Detect duplicate author initials
@@ -31,7 +31,7 @@ def get_author_info(path):
         raise ValueError(msg)
 
     # Replace missing values with None
-    author_df.fillna("None", inplace=True)
+    #author_df.fillna("None", inplace=True)
 
     authors = author_df.to_dict(orient='records')
     return authors
--- a/build/authors.py
+++ b/build/authors.py
@@ -31,7 +31,7 @@ def get_author_info(path):
         raise ValueError(msg)
 
     # Replace missing values with None
-    author_df.where((pandas.notnull(author_df)), None,inplace=True)
+    author_df.where(pandas.notnull(author_df), None, inplace=True)
 
     authors = author_df.to_dict(orient='records')
     return authors
--- a/build/authors.py
+++ b/build/authors.py
@@ -3,7 +3,6 @@ Process author table
 """
 
 import pandas
-import numpy
 
 
 def get_author_info(path):
--- a/build/authors.py
+++ b/build/authors.py
@@ -3,6 +3,7 @@ Process author table
 """
 
 import pandas
+import numpy
 
 
 def get_author_info(path):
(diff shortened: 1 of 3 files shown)


build/authors.py
```diff
@@ -3,6 +3,7 @@
 """
 
 import pandas
+import numpy
```
dhimmel: No need for numpy. These checks can be done in pandas.

build/authors.py
```diff
@@ -11,12 +12,27 @@ def get_author_info(path):
...
+
     # Detect duplicate author initials
     duplicated_initials = list(
         author_df[author_df.duplicated('initials')].initials)
     if duplicated_initials:
         msg = f'Duplicated initials in authors.tsv: {duplicated_initials}'
         raise ValueError(msg)
 
+    # Replace missing values with None
+    author_df.fillna("None", inplace=True)
```
dhimmel: What's the issue with keeping as missing or `None`? I don't like the possibility of string collisions.

Does `{% if author.orcid -%}` or `{% if not none(author.orcid) -%}` not work in jinja2?

content/00.front-matter.md
```diff
@@ -4,9 +4,16 @@
 
 {% for author in authors %}
 + **{{author.full_name}}**<br>
-  [\@{{author.github_username}}](https://github.com/{{author.github_username}}) ·
-  [{{author.orcid}}](https://orcid.org/{{author.orcid}}) ·
-  {{author.affiliations}} ·
-  Funded by {{author.funders}}
+  {% if author.github_username != 'None' -%}
+    [\@{{author.github_username}}](https://github.com/{{author.github_username}}) ·
```
dhimmel: Should the ` · ` be prepended to included items. Right now you can get trailing interpuncts?

content/00.front-matter.md
```diff
@@ -4,9 +4,16 @@
...
+  {%- endif %}
+  {% if author.orcid != 'None' -%}
```
dhimmel: `-%}` what does `-` do there? I've seen it in some places... is it needed?

vsmalladi: -%} Preserves spacing by not introducing new lines

build/authors.py
vsmalladi: Fixing to only be done with pandas

build/authors.py
vsmalladi: {% if author.orcid -%} : Checks if author.orcid has a value so this is always True

{% if not none(author.orcid) -%}: Was not working as nan was not check as None. So changed the code that that NaN is replaced as None in pandas rather than string 'None'

content/00.front-matter.md
```diff
@@ -4,9 +4,21 @@
 
 {% for author in authors %}
 + **{{author.full_name}}**<br>
-  [\@{{author.github_username}}](https://github.com/{{author.github_username}}) ·
-  [{{author.orcid}}](https://orcid.org/{{author.orcid}}) ·
-  {{author.affiliations}} ·
-  Funded by {{author.funders}}
+  {% set previous = false %}
```
agitter: The `previous` flag makes this code block more complicated.  Could we simplify the code if we put some required information first, such as the affiliation?  That would also involve making an affiliation required in addition to the name and initials.

@dhimmel what do you think?

content/authors.tsv
```diff
@@ -1,3 +1,4 @@
 github_username	full_name	initials	orcid	email	affiliations	funders
 dhimmel	Daniel S. Himmelstein	DSH	0000-0002-3012-7446	daniel.himmelstein@gmail.com	Department of Systems Pharmacology and Translational Therapeutics, University of Pennsylvania	GBMF4552
 agitter	Anthony Gitter	AG	0000-0002-5324-9833	gitter@biostat.wisc.edu	Department of Biostatistics and Medical Informatics, University of Wisconsin-Madison and Morgridge Institute for Research	NIH U54AI117924
+vsmalladi	Venkat S. Malladi	VSM	0000-0002-0144-0564	vsmalladi@gmail.com	The Laboratory of Signaling and Gene Expression, Cecil H. and Ida Green Center for Reproductive Biology Sciences, University of Texas Southwestern Medical Center
```
agitter: I don't care either way, but I don't think there is much incentive for other new contributors to add themselves to the `authors.tsv` list.  We should rely on https://github.com/greenelab/manubot-rootstock/graphs/contributors to see who contributed.  I'm even fine with my name being removed as long as we have 2-3 example authors.

content/00.front-matter.md
vsmalladi: @agitter Thats possible, I don't have any preferences.

content/authors.tsv
vsmalladi: @agitter I just put myself in as an example for missing information and how it would look. I can remove myself too and just put some dummy authors.

agitter: @vsmalladi I think it's fine having 3 examples.  I made the comment in part so that other contributors don't think it is necessary.

vsmalladi: @agitter Sounds great.

content/00.front-matter.md
dhimmel: Maybe just assume (but do not enforce) that GitHub username will always be there. A leading interpunct is not the end of the world. Or you could switch to ORCID first and assume its always present. This way you'd incent everyone to have an ORCID.

vsmalladi: @dhimmel That sounds reasonable. I can make the ORCID as first but not enforce it. This will simplify the code.

build/authors.py
```diff
@@ -11,12 +11,27 @@ def get_author_info(path):
...
+
     # Detect duplicate author initials
     duplicated_initials = list(
         author_df[author_df.duplicated('initials')].initials)
     if duplicated_initials:
         msg = f'Duplicated initials in authors.tsv: {duplicated_initials}'
         raise ValueError(msg)
 
+    # Replace missing values with None
+    author_df.where((pandas.notnull(author_df)), None,inplace=True)
```
dhimmel: extra parentheses and missing space after `None,`.

Issue #14, opened 2017-07-11
SVG images fail to export to PDF


dhimmel:
Refs https://github.com/jgm/pandoc/issues/265

slochower:
That thread seems to indicate a problem going through LaTeX, but I think the problem persists if we do Markdown → HTML5 → PDF as is currently implemented. Are you considering creating a local rasterized copy of the graphics to generate the PDF? Something like this may be helpful: https://gist.github.com/jeromerobert/3996eca3acd12e4c3d40

dhimmel:
> Markdown → HTML5 → PDF

Yes, I agree that the problem is occurring for this lineage.

> Are you considering creating a local rasterized copy of the graphics to generate the PDF?

I think there should be a solution that doesn't require rasterizing vectors. For example, if an SVG is converted to PDF, then it should be able to be embedded in a PDF without sacrificing the vectors?

slochower:
> For example, if an SVG is converted to PDF, then it should be able to be embedded in a PDF without sacrificing the vectors?

I hope so. Can you make a PDF version of the SVG file we're using for the figure example, built from the same source as the SVG?

vsmalladi:
Should there be a local directory for the figures in the repository?

dhimmel:
@vsmalladi see discussion on https://github.com/greenelab/manubot-rootstock/issues/1: currently we're focusing on images via URL. In #1, @slochower has a working solution for images via paths, but we didn't see a pressing need that justified the complexity.

vsmalladi:
@dhimmel thanks for the brining me up to speed.

vsmalladi:
Accidentally
 put this in the wrong place: [index.pdf](https://github.com/greenelab/manubot-rootstock/files/1185668/index.pdf)


I have been playing with the using wkhtmltopdf and converting the images to png to get them to render for the pdf, based on code in #40. I have attached an example of converting from markdown to html5.

agitter:
@vsmalladi The figures show up, so that's a big improvement.  All text and figures are tiny when I view the pdf though.  Do you see that as well?

vsmalladi:
@agitter Ya I had the same issue. Wanted to see if anyone else saw the same thing. Not sure what is happening with the scaling. But it is looking better.

dhimmel:
[...]

PR #18, opened 2017-07-15
Use pandoc_yaml_metadata and remove authors.tsv

dhimmel:
Follows up on and closes https://github.com/greenelab/manubot-rootstock/pull/16.


dhimmel:
@agitter one issue with defining `author` in a `yaml_metadata_block` is that pandoc adds authors to both the document metadata and document body. The way it adds authors to the document body is ugly, but I don't think there is away to suppress it. Therefore for HTML I used sed to delete the h2 author lines. For PDF they remain in the text: see [manuscript.pdf](https://github.com/greenelab/manubot-rootstock/files/1150437/manuscript.pdf).

So if we merge this our PDF becomes even more ugly than it already is. Personally I plan to create PDFs by printing the HTML page to PDF. But @agitter what are your thoughts here?

dhimmel:
Here's the metadata that's added to HTML:

```html
  <meta name="author" content="Daniel S. Himmelstein" />
  <meta name="author" content="Anthony Gitter" />
  <meta name="author" content="Venkat S. Malladi" />
  <meta name="author" content="Evan M. Cofer" />
  <meta name="keywords" content="work-in-progress, markdown, manuscript, publishing" />
  <title>Manubot Rootstock: Manuscript Title</title>
```

Unfortunately, PDF metadata doesn't seem to have transmitted:

pdf-metadata

agitter:
Between this and #14, we may need to make a call on how important PDF is to us.  I think PDF is still much more standard for a manuscript than HTML, especially for saving locally, so I'm hesitant to abandon it completely.  But are we willing to have ugly PDFs?  Or live with a LaTeX dependency because [that will add the metadata to the PDF](https://github.com/greenelab/manubot-rootstock/pull/16#issuecomment-314629066)?

As a partial workaround, we could leave the author list in both the HTML and PDF and then move the detailed `Authors` section to the end of the manuscript.  I'm not a huge fan of that because it makes it harder to contact the authors, and some readers probably value having affiliations up front more than I do.

I wish I knew pandoc better so I could help come up with LaTeX-free PDF solutions.

dhimmel:
> I think PDF is still much more standard for a manuscript than HTML

In Chrome, print to PDF does a great job converting the manuscript to PDF. See how the [Sci-Hub manuscript](https://greenelab.github.io/scihub-manuscript/) converts to [`Sci-Hub-Chrome-Print-to-File.pdf`](https://github.com/greenelab/manubot-rootstock/files/1150648/Sci-Hub-Chrome-Print-to-File.pdf). So the question is whether we can script this conversion. `wkhtmltopdf` is supposed to do this conversion, but I'm not sure why it's performing poorly when used with pandoc.

If I run the following locally:

```
wkhtmltopdf https://greenelab.github.io/scihub-manuscript/ Sci-Hub-wkhtmltopdf.pdf
```

I get [Sci-Hub-wkhtmltopdf.pdf](https://github.com/greenelab/manubot-rootstock/files/1150652/Sci-Hub-wkhtmltopdf.pdf), which is better than the pandoc version, but not as good as the Chrome version.

agitter:
The Chrome-generated version does look good.  I like having the versioned PDFs on the `gh-pages` branch because I don't want to assume every reader has Chrome available.  Is it too confusing if we recommend that readers print the PDF through Chrome and fall back to the PDF we built only as a secondary option?

An immediate build improvement could be to switch to calling `wkhtmltopdf` directly instead of `pandoc` for the PDF generation.  Because `wkhtmltopdf` is available in the conda environment, can we do something like `wkhtmltopdf output/index.html output=output/manuscript.pdf`?  That can follow your `sed` call so that the extra author list is removed.

`wkhtmltopdf` isn't perfect - the first line of the SciHub manuscript intro overlaps reference `[1]` and `literature` - but that might be enough for us to merge.  

Then we can open another issue or use #4 to consider better ways to build the PDF.  My impression is that we'll need `pdflatex` to automatically build high-quality PDFs.  I haven't looked through these, but there is a remote chance we could [install it through `conda`](https://anaconda.org/search?q=texlive).  Others may be using [texlive-core](https://anaconda.org/conda-forge/texlive-core) for [purposes similar to ours](https://github.com/conda-forge/texlive-core-feedstock/issues/14).  I haven't tested it.

dhimmel:
> An immediate build improvement could be to switch to calling wkhtmltopdf directly instead of pandoc for the PDF generation

Done in https://github.com/greenelab/manubot-rootstock/pull/18/commits/8d8d134f4d08cbc11ea34a067bb6a37d391cb2b3. I agree this is a major improvement, since we no longer have to worry about multiple pandoc conversions succeeding. We can just worry about the HTML, do any post-processing, and then essentially print to PDF.

> Is it too confusing if we recommend that readers print the PDF through Chrome and fall back to the PDF we built only as a secondary option?

I think this is a good idea. Let's see where we get in a few weeks and if we haven't improved things, we'll add this note. Since Chromium is open source, there is a possibility that we could access their print preview system via an API (may be difficult as per https://github.com/electron/electron/issues/805).

> wkhtmltopdf isn't perfect - the first line of the SciHub manuscript intro overlaps reference [1] and literature - but that might be enough for us to merge.

Calling `wkhtmltopdf` directly doesn't fix the svg failures of #14. But atleast, we now know the problem lies with wkhtmltopdf and there may be workarounds.

slochower:
> Done in 8d8d134. I agree this is a major improvement, since we no longer have to worry about multiple pandoc conversions succeeding. We can just worry about the HTML, do any post-processing, and then essentially print to PDF.

Do know if this would work even if using something like the javascript library KaTeX for better math support in the HTML output?

dhimmel:
> Do you know if this would work even if using something like the javascript library KaTeX for better math support in the HTML output?

`wkhtmltopdf` appears to support javascript modifications. Let's have future discussion on this topic at #20

PR #21, opened 2017-07-16
Add GitHub, ORCID, Twitter icons

dhimmel:
Adds images directory as pioneered by @slochower in greenelab#8.

Uses symbolic link so content/images maps to output/images.

Fix https://twitter.com/gcabanac/status/886587167333642240

Makes affiliations & funding small text.


dhimmel:
Rendered HTML now looks like

authorship

@agitter what do you think? Do you have a preference for or against making affiliation text `<small>` like in the screenshot above?

content/00.front-matter.md
```diff
@@ -2,14 +2,22 @@
 
 {% for author in authors %}
 + **{{author.full_name}}**<br>
+  ![ORCID icon](images/orcid.svg){height="13px"}
```
agitter: Adding the icon makes the soft requirement that all authors have an ORCID stronger.  The icon looks strange without the id.

I'm okay with that, but let's add `0000-0003-3877-0433` for @evancofer to `metadata.yaml`.

content/00.front-matter.md
```diff
@@ -2,14 +2,22 @@
...
   {%- if author.github_username is defined %}
-     · [\@{{author.github_username}}](https://github.com/{{author.github_username}})
+    · ![GitHub icon](images/github.svg){height="13px"}
+    [{{author.github_username}}](https://github.com/{{author.github_username}})
   {%- endif %}
+  {%- if author.twitter_username is defined %}
+    · ![Twitter icon](images/twitter.svg){height="13px"}
+    [{{author.twitter_username}}](https://twitter.com/{{author.twitter_username}})
+  {%- endif %}<br>
+  <small>
```
agitter: The `<small>` affiliation text looks better.

dhimmel:
> Does #14 affect the icons in the PDF? It shouldn't be critical if they are lost, and we need a general solution for that anyway.

Correct, icons do not get exported to the PDF.

> Are we now changing the decision to support local images (#8), or is this an exception to the general figure handling?

Local images should now work. Where do you think the best place to document image behavior is? In the manuscript or in a README?

content/00.front-matter.md
dhimmel: We could also check whether orcid is defined and not print the symbol? This would leave only a preceeding interpunct which isn't too big of a deal. Or we could print `missing ORCID` next to the symbol if we really want to nudge authors to use ORCID.

vsmalladi:
@dhimmel I think Readme

content/00.front-matter.md
agitter: Let's only print the symbol if the ORCID is defined.

agitter:
> Readme would be good about generals where to put images and manuscript for examples

I agree with this.  Once they start writing a new manuscript in a cloned repo, a user will remove the example `.md` files.  It would be for them to still have instructions in their repo instead of having to come back to this repo.

We could consider adding a new `.md` file with usage instructions.  Currently it is split across `README.md`, `references/README.md`, and `CONTRIBUTING.md`.

PR #24, opened 2017-07-16
Write source repo version info in manucript

dhimmel:
Only applies to manuscripts built on Travis. Creates a header under the title with version information. Therefore, given a manuscript that's built via Travis, you can always get back to the source:

commit-in-manuscript

Also will help alert readers to the online source and manubot system.


agitter:
Looks good to me

PR #26, opened 2017-07-16
Convert to docx

vsmalladi:
[manuscript.docx](https://github.com/greenelab/manubot-rootstock/files/1151389/manuscript.docx)
 
I have attached an example of the word document you can get with some performing.

Issues:

- SVG embedding like PDF #14. Possible work around filter [pandoc-svg.py](https://gist.github.com/jeromerobert/3996eca3acd12e4c3d40)

- Equation numbering doesn't work due to known formatting  issue with word [pandoc-eqnos #16](https://github.com/tomduck/pandoc-eqnos/issues/16)


vsmalladi:
@dhimmel I am working on trying to use local images as well. But run into this error
`[pandoc warning] Could not find image `images/logo.svg', skipping...`

How did you resolve this error?

build/build.sh
```diff
@@ -43,3 +44,20 @@ wkhtmltopdf \
   --quiet \
   output/index.html \
   output/manuscript.pdf
+
+echo "Exporting Word Docx manuscript"
+pandoc --verbose \
+--smart \
```
dhimmel: I think `--smart` is most likely a good thing for us... but let's try to keep in sync with HTML output as much as possible. So also add to HTML command or save for a future PR.

build/build.sh
```diff
@@ -43,3 +44,20 @@ wkhtmltopdf \
...
+--from=markdown \
+--to=docx \
+--filter pandoc-fignos \
+--filter pandoc-tablenos \
+--bibliography=$BIBLIOGRAPHY_PATH \
+--csl=$CSL_PATH \
+--reference-docx=$DOCX_PATH \
```
dhimmel: Can you explain what this is? What happens if you don't pass it?

build/build.sh
```diff
@@ -43,3 +44,20 @@ wkhtmltopdf \
...
+--self-contained \
+--metadata link-citations=true \
+--include-in-header=build/assets/analytics.js \
+--include-after-body=build/assets/anchors.js \
```
dhimmel: I suspect both these javascript files only affect HTML output so they can be removed here.

build/build.sh
```diff
@@ -43,3 +44,20 @@ wkhtmltopdf \
   --quiet \
   output/index.html \
   output/manuscript.pdf
+
+echo "Exporting Word Docx manuscript"
```
dhimmel: I'm not sure we want the word export as part of every build. Primarily do we want to track these files on GitHub which will add to repo size? Note you can always use git to go back to a certain point in the repos history and generate one of these docs.

Do we instead want to make it optional via some sort of argument / environment variable? Or even put in a different script?

agitter: I like the idea of providing a pandoc command that generates docx if we can get a reasonable docx version.  I don't have a strong opinion about where that command should go.

If we don't track the docx files in the repo, a user will have to be more computationally savvy to use them.  It's easy for us to run the build script from a particular point in the history, but less-computational users may not be able to.  I'm not sure whether to tradeoff storage for ease of export until we know the target audience better.

agitter:
The table and figure links work correctly for me in the docx, but the reference links are broken.

build/build.sh
vsmalladi: > Can you explain what this is? What happens if you don't pass it?
not passing `--reference-docx=$DOCX_PATH` makes some default formatting options that the user would change. I think it uses the Normal template in Word. Thus font in Calabri font instead of Times New Roman. Some of the Headers are Blue instead of Black. Its not necessary, but users can change the style settings and the next time they generate the word doc it will use those styles. Allowing users to change the styles necessary for different journals or collaborators.

build/build.sh
vsmalladi: @dhimmel Ya I can remove this.

build/build.sh
vsmalladi: @dhimmel I can add this to the HTML command as well.

build/build.sh
vsmalladi: I like the idea of providing a command option that generates the docx.

vsmalladi:
> Looks like you need to rebase on the current version of this repo. Try not to start PRs on top of commits that are not part of the upstream master (as appears to have happened here)

Thanks for the catch. I will fix this PR.

vsmalladi:
@dhimmel Looks like I used merge instead of rebase with the following commands on my master to bring it up to speed:
`git fetch upstream
git merge upstream/master`

Any way to fix this with rebase?

dhimmel:
> Any way to fix this with rebase?

Yeah but using `git rebase upstream/master`. Since your diff looks fine, this is not essential, but it will get rid of the superfluous commits. In the end we'll squash all your commits, so this really only matters for PR review.

vsmalladi:
> Yeah but using git rebase upstream/master. Since your diff looks fine, this is not essential, but it will get rid of the superfluous commits. In the end we'll squash all your commits, so this really only matters for PR review.

I will keep this is mind for future features.

vsmalladi:
> but the reference links are broken.

@agitter The link citation seems to be fixed for a future release of pandoc's: [pandoc-citeproc #205](https://github.com/jgm/pandoc-citeproc/issues/205)

build/build.sh
dhimmel: >  I'm not sure whether to tradeoff storage for ease of export until we know the target audience better.

What about an environment variable `BUILD_DOCX`? Then `if [ $BUILD_DOCX = "true" ]`, we build the DOCX. Therefore users can activate the feature by running

```
BUILD_DOCX=true sh build/build.sh
```

Or setting `BUILD_DOCX=true` via their Travis configuration or options.

vsmalladi: > What about an environment variable BUILD_DOCX? Then if [ $BUILD_DOCX = "true" ], we build the DOCX. Therefore users can activate the feature by running

I like this idea. Gives the flexibility if they want to not build the DOCX every time.

agitter: > environment variable BUILD_DOCX

I like this as well.

build/build.sh
```diff
@@ -9,6 +9,7 @@ echo "Retrieving and processing reference metadata"
 
 # pandoc settings
 CSL_PATH=references/style.csl
+DOCX_PATH=references/reference-docx.docx
```
dhimmel: Can we change this filename to `pandoc-reference.docx`? I think that will make it a little more clear what it's purpose is.

build/build.sh
```diff
@@ -43,3 +45,21 @@ wkhtmltopdf \
...
+    pandoc --verbose \
+    --smart \
+    --from=markdown \
+    --to=docx \
+    --filter pandoc-fignos \
+    --filter pandoc-tablenos \
+    --bibliography=$BIBLIOGRAPHY_PATH \
+    --csl=$CSL_PATH \
+    --reference-docx=$DOCX_PATH \
+    --metadata link-citations=true \
```
dhimmel: Note that we now have the option to put `link-citations: true` in `metadata.yaml`. Not sure if we should do this... if so we can do this in a separate PR.

build/README.md
```diff
@@ -2,6 +2,7 @@
 
 [`build.sh`](build.sh) builds the repository.
 `sh build.sh` should be executed from the root directory of the repository.
+`BUILD_DOCX=true sh build/build.sh` should be executed to build DOCX document, otherwise will be skipped.
```
dhimmel: Can we expand the doc to something like:

```md
To build a DOCX file of the manuscript, set the `BUILD_DOCX` environment variable to `true`.
For example, use the command `BUILD_DOCX=true sh build/build.sh`.
Set a [Travis environment variables](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings) to export DOCX for all Travis builds.
Currently, equation numbers via `pandoc-eqnos` are not supported for DOCX output.
```

build/build.sh
vsmalladi: @dhimmel I like the idea. Will work on updating the file.

build/build.sh
vsmalladi: @dhimmel I can remove this line:
` --metadata link-citations=true`

build/README.md
vsmalladi: Sounds much better.

dhimmel: Make sure to remove my extra s:

> Set a Travis environment variable**s**

dhimmel:
link-citations is not set in metadata.yaml? Is it?

vsmalladi:
>Note that we now have the option to put `link-citations: true` in metadata.yaml. Not sure if we should do this... if so we can do this in a separate PR.

@dhimmel I misinterpreted this and revert that change. We can change it in new PR for HTML and DOCX.

PR #35, opened 2017-07-20
page-break control using HTML classes

dhimmel:
Works for both print output and wkhtmltopdf.


dhimmel:
I'm going to merge cause I want to use this feature downstream immediately, but revisions welcome.

agitter:
Nice feature.  What elements are supported?  Is it primarily to be used alongside markdown headers?

dhimmel:
Haven't tested it out too extensively, but any HTML element where pandoc accepts the `{.page_break_before}` syntax for should work. I tried adding `.page_break_inside_avoid` to image attributes to force the caption to be on the same page as the image, but the class only ended up being applied to the image and not the image+caption element.

agitter:
For the example manuscript, would it work to add `{.page_break_after}` to the abstract instead of `{.page_break_before}` to the equations?  I expect that is one of the most common uses, and it will persist even if someone deletes `02.equations.md`.

dhimmel:
> would it work to add {.page_break_after} to the abstract instead of {.page_break_before} to the equations?

That doesn't appear to work, but the following does appear to work:

```
<div class="page_break_after"></div>
```

agitter:
Thanks for testing.  We can leave it as is in that case.  I think that HTML in the markdown files should be avoided as much as possible.

PR #36, opened 2017-07-20
Add date to output

dhimmel:
This adds a date in both the HTML `<head>` like:

```html
<meta name="date" content="2017-07-20">
```

And prints the date in the output like either:

> This manuscript was automatically generated on July 20, 2017.

Or if via a Travis master build:

> This manuscript was automatically generated from greenelab/manubot-rootstock@xxXXxxX on July 20, 2017.


vsmalladi:
Looks great.

PR #39, opened 2017-07-21
Set print margin-top and margin-bottom

dhimmel:
Closes https://github.com/greenelab/manubot-rootstock/issues/38

With this PR, here's what the PDFs look like:

+ generated in `build.sh`: [manuscript-wkhtmltopdf.pdf](https://github.com/greenelab/manubot-rootstock/files/1166592/manuscript-wkhtmltopdf.pdf)
+ print preview in chrome on linux [manuscript-chrome-linux.pdf](https://github.com/greenelab/manubot-rootstock/files/1166598/manuscript-chrome-linux.pdf)

Note that I added a section (not tracked) with lorem ipsum text.

Also `--print-media-type` gets rid of border from the `wkhtmltopdf` PDF.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -40,8 +40,10 @@ echo "Exporting PDF manuscript"
 wkhtmltopdf \
   --quiet \
   --print-media-type \
-  --margin-top 22 \
-  --margin-bottom 20 \
+  --margin-top 21 \
+  --margin-bottom 17 \
+  --margin-left 0 \
+  --margin-right 0 \
   output/index.html \
   output/manuscript.pdf
 
--- a/build/build.sh
+++ b/build/build.sh
@@ -39,6 +39,9 @@ pandoc --verbose \
 echo "Exporting PDF manuscript"
 wkhtmltopdf \
   --quiet \
+  --print-media-type \
+  --margin-top 22 \
+  --margin-bottom 20 \
   output/index.html \
   output/manuscript.pdf
 
--- a/output/github-pandoc.css
+++ b/output/github-pandoc.css
@@ -77,10 +77,7 @@ body {
     margin: 0;
 }
 
-/* 
- * Page margins when printing
- * https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2982#issuecomment-308714479
- */
+/***Page margins when printing***/
 @page { 
     margin-top: 22mm;
     margin-bottom: 20mm;
--- a/output/github-pandoc.css
+++ b/output/github-pandoc.css
@@ -79,8 +79,10 @@ body {
 
 /***Page margins when printing***/
 @page { 
-    margin-top: 22mm;
-    margin-bottom: 20mm;
+    margin-top: 19mm;
+    margin-bottom: 16mm;
+    margin-left: 0mm;
+    margin-right: 0mm;
 }
 
 /* 
--- a/output/github-pandoc.css
+++ b/output/github-pandoc.css
@@ -77,13 +77,21 @@ body {
     margin: 0;
 }
 
+/* 
+ * Page margins when printing
+ * https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2982#issuecomment-308714479
+ */
+@page { 
+    margin-top: 22mm;
+    margin-bottom: 20mm;
+}
 
 /* 
  * Page Breaks
  * https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2982#issuecomment-308714479
  */
 
-@media screen, print {
+@media print {
   /***Always insert a page break before the element***/
   .page_break_before {
       page-break-before: always !important;


output/github-pandoc.css
```diff
@@ -77,13 +77,21 @@ body {
     margin: 0;
 }
 
+/* 
+ * Page margins when printing
+ * https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2982#issuecomment-308714479
```
agitter: Is this URL relevant here?  I didn't see margins discussed in that issue.

build/build.sh
```diff
@@ -39,6 +39,9 @@ pandoc --verbose \
 echo "Exporting PDF manuscript"
 wkhtmltopdf \
   --quiet \
+  --print-media-type \
+  --margin-top 22 \
+  --margin-bottom 20 \
```
agitter: Is 22 and 20 a standard format, or just what happens to look good?

output/github-pandoc.css
dhimmel: Good catch! Accidentally included from a copy and paste. Fixed in https://github.com/greenelab/manubot-rootstock/pull/39/commits/3ee2d67b82d99e2cb0dc41ecb87e5a8522a67dcb

dhimmel:
> I prefer the smaller margins when there isn't a header added by a preprint server

Agreed. Especially since we don't have floating images / tables. Since images / tables are wherever they're specified, there can be lot's of wasted space. Minimal margins help prevent big blank areas. Given that many journal systems will apply headers and footers, I think top and bottom margin are more important. I'm okay with smaller side margins, but when text spans too wide, it's harder to read... it becomes hard to always navigate to the next line.

> it's easiest if the default margins are compatible with PeerJ and others.

I used the PeerJ guidelines as a starting point. From the [PeerJ Guidelines](https://peerj.com/about/author-instructions/)

> Include line numbers, an approx. 2.5cm margin on all edges, and 12 point Times font for readability. This makes your review pdf easy to read, and your editor and reviewers very happy!

One thing to note is that I believe some additional margin is getting added beyond what's specified in `margin-top` or `margin-bottom` or `margin-right` or `margin-left`. Maybe this is coming from somewhere else in the CSS. I didn't fully diagnose how the magins worked. Just played around until things look reasonable in both `wkhtmltopdf` and print to PDF. I tried to make it so `wkhtmltopdf` and print to PDF have the same margins.

Probably the easiest solution is to look at the PDFs I posted and decide where we want to decrease those margins... then play around until we find the right settings.

dhimmel:
I played around some more and took some actual measurements of margins on printed papers. As of https://github.com/greenelab/manubot-rootstock/pull/39/commits/50e444dc06ba3334e88797c499374e10818e6f5d, here are the PDFs for the two methods of export:

1. [manuscript-wkhtmltopdf-v2.pdf](https://github.com/greenelab/manubot-rootstock/files/1170493/manuscript-wkhtmltopdf-v2.pdf) created via `wkhtmltopdf` in `build.sh`. Top margin 24mm, bottom margin 21mm, and sides of 32mm.

2. [manuscript-chrome-linux-v2.pdf](https://github.com/greenelab/manubot-rootstock/files/1170494/manuscript-chrome-linux-v2.pdf) created using print to PDF on Chrome for Linux. Top margin of 24mm, bottom of 24mm, sides of 29mm.

I also noticed that Chrome's print to PDF has a custom margin option that allows you to override the default margins. In instances where margins really matter (like submitting to a journal), authors may want to tweak the margins to perfection using this method. At least until we get a frontend contributor that can make everything harmoniously just work.

agitter:
Someone reminded me today that the deep review is 100+ pages when printed to PDF from HTML.  How hard would it be to port these margin fixes back into that repository?  Should we update the css manually instead of trying to merge in these commits?

dhimmel:
> How hard would it be to port these margin fixes back into that repository?

I think I should be able to do it with an advanced git merge. After this first merge, future merges should be easier. Lately it hasn't been too hard to merge changes back into `scihub-manuscript`.

agitter:
@dhimmel Thinking about this more, I realized that this pull request actually increased the top and bottom margin (though decreasing left and right).  I was proposing to decrease all margins in deep review.  We could merge this pull request and then customize the deep review margins if needed.

This is not time sensitive.  If you prefer to sync with deep review later that's fine with me.

dhimmel:
>  I realized that this pull request actually increased the top and bottom margin (though decreasing left and right). I was proposing to decrease all margins in deep review.

The Deep Review PDF does not get any CSS styling applied to it. The overall effect of merging the current `manubot-rootstock` into `deep-review` would likely be to increase the text per page.

The big issue will be the author formatting changes. @agitter did you want to retain the deep-review author system, or would you be interested in switching to this system?

agitter:
Because deep review does more processing with the authors to generate the funding statement and affiliation numbers, I prefer to keep it as is.  What do you think?

dhimmel:
> I prefer to keep it as is. What do you think?

I think keeping the current authorship format of Deep Review makes sense. However, it will make resolving the differences between manubot and deep-review more difficult. That's fine, but perhaps we should hold off on the merge then. I'm still not entirely sure the manubot authorship framework is entirely stable yet.

And I want to port lot's of the python code to the `manubot` package. This will likely cause drastic architecture changes.

But if you think the deep-review is suffering by having an out of date build system, we can do the merge sooner.

agitter:
I'm reconsidering after our discussion of how deep review and manubot rootstock have diverged.  Let's delay a full merge.  We could manually modify `github-pandoc.css` in deep review, but if that will further complicate a merge later it isn't critical.

PR #62, opened 2017-08-28
WIP: Documentation and cloning improvements

dhimmel:
Work in progress


dhimmel:
@slochower @zietzm @agitter: https://github.com/greenelab/manubot-rootstock/pull/62/commits/6b56fe607720ae2f0cee6fbd86f5c59641891041 links to instances of the manubot. I added some of your work. If you'd rather hold off on publicizing these repos, let me know, and I will remove them. Can respond privately to `daniel.himmelstein@gmial.com` if desired.

slochower:
I'd prefer not for the specific repository of mine you've linked because that document is part of a draft that's currently making the rounds among coauthors at the moment, but we can do another document.

SETUP.md
```diff
@@ -0,0 +1,166 @@
...
+
+```sh
+# GitHub account (change from greenelab)
+OWNER=greenelab
+# Repository name (change from manubot-rootstock)
+REPO=manubot-rootstock
+```
+
+The following commands assume you have SSH key access configured for the `OWNER` GitHub account.
+If this is not the case, see GitHub's guidelines for [generating](https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/) a new SSH key and [adding it](https://help.github.com/articles/adding-a-new-ssh-key-to-your-github-account/) to your account.
```
agitter: We could make this step optional, but the user would have to know how to modify some of the git commands below.

SETUP.md
```diff
@@ -0,0 +1,166 @@
...
+
+```sh
+git push --set-upstream origin master
+git push --set-upstream origin gh-pages
+git push --set-upstream origin output
+```
+
+## Continuous integration
+
+Now you must manually enable Travis CI for the new repository at https://travis-ci.org.
```
agitter: Could link to https://docs.travis-ci.com/user/getting-started/ for users who aren't already using Travis CI.

USAGE.md
```diff
@@ -140,3 +140,15 @@ funders: GBMF4552  # optional
...
 If you experience any issues with the Manubot or would like to contribute to its source code, please visit [`greenelab/manubot`](https://github.com/greenelab/manubot) or [`greenelab/manubot-rootstock`](https://github.com/greenelab/manubot-rootstock).
+
+## Examples
+
+For additional examples, check out existing manuscripts that use the Manubot:
+
++ The Sci-Hub Coverage Study ([source](https://github.com/greenelab/scihub-manuscript), [manuscript](https://greenelab.github.io/scihub-manuscript/))
++ A Report for the Vagelos Scholars Program by Michael Zietz ([source](https://github.com/zietzm/Vagelos2017), [manuscript](https://zietzm.github.io/Vagelos2017/))
++ The Deep Review ([source](https://github.com/greenelab/deep-review), [manuscript](https://greenelab.github.io/deep-review/))
++ The Meta Review ([source](https://github.com/greenelab/meta-review), [manuscript](https://greenelab.github.io/meta-review/))
```
agitter: I think it is fine to link the meta review here because it's already public.  We can update the name once we finalize it.  We should confirm with @cgreene as well that this is okay to share.

cgreene: Yes - fine with me :+1:

SETUP.md
dhimmel: I think that link may be more confusing than the landing page for users who haven't authorized a GitHub account at https://travis-ci.org.

dhimmel:
@slochower what about `synthetic-motor-literature` as per https://github.com/greenelab/manubot-rootstock/pull/62/commits/d2d382f5d7b9bdce88be63f1eb62f3f746991ebc?

dhimmel:
@agitter to prevent conflicts when updating to the latest upstream rootstock, we should clearly delineate files or subsections of files that users edit. Then as long as users only touch these sections and we freeze upstream changes on these sections, there shouldn't be conflicts.

This is part of the reason why I deleted several of the markdown content files. But I'm tempted to go further and perhaps delete every md file besides the front and back matters. Maybe also keep the blank abstract page?

Potentially clean out `metadata.yaml`, so it's mostly blank? Or change it so it's more documentation. @agitter @slochower: what do you think?

agitter:
I definitely agree with should minimize conflicts and lock certain files.   What will happen with `README.md` or other files that are edited in both the cloned repo and `manubot-rootstock`?  Will we tell users to avoid merging them when they pull updates?

We may need to decide to prioritize novice users who want more examples in the `manubot-rootstock` repository or advanced users who want to merge upstream changes.  We could make a new `manubot-example` repo to provide a complete working example, but that still seems less accessible than a minimal example in the cloned repo that tells the user what example to modify.

In a way, this reminds me of a LaTeX template from a publisher.  Those usually have example text to demonstrate the style instead of a blank slate as the starting point.

dhimmel:
> What will happen with README.md or other files that are edited in both the cloned repo and manubot-rootstock?

Editing the same file shouldn't generally be an issue. Editing the same lines is an issue. A conflict will occur and the user will have to resolve the conflict to continue the merge process.

> In a way, this reminds me of a LaTeX template from a publisher. Those usually have example text to demonstrate the style instead of a blank slate as the starting point.

Good point. Okay let's leave some template material in content. We should just be careful not to change the template content in the future (since that will create conflicts).

slochower:
@dhimmel Sure.

Is the template material going to be in `content/`? If the users pull from `greenelab/manubot-rootstock` won't that create duplicate sections `01...md`, `02...md`, and so on?

agitter:
> If the users pull from greenelab/manubot-rootstock won't that create duplicate sections 01...md, 02...md, and so on?

I was wondering the same thing.  If we don't modify `content/02.delete-me.md` in `manubot-rootstock` and a user deletes it in their clone, then I expect git would not not restore the file during the merge.  I'm not confident about that behavior or what would happen if we do make modifications upstream.

dhimmel:
> If the users pull from greenelab/manubot-rootstock won't that create duplicate sections `01...md`, `02...md`, and so on?

That's why I want to minimize the content md files in this repo.

> If we don't modify `content/02.delete-me.md` in manubot-rootstock and a user deletes it in their clone, then I expect git would not not restore the file during the merge.

Correct.

Now it may have a conflict if the user deletes `02.delete-me.md` and then rootstock modifies `02.delete-me.md`. However, this conflict doesn't require much thinking to resolve... just keep the file deleted.

dhimmel:
https://github.com/greenelab/manubot-rootstock/pull/62/commits/e83ab6d9ac5e865a9ccc0bf06412db1322f22f6f changes authorship in `metadata.yaml` to John Doe / Jane Roe. The goal is to make it clear that these are just placeholder authors and not the actual entire authorship for:

1. the Manubot project
2. freshly cloned projects that haven't updated `metadata.yaml` yet.

agitter:
I like the switch to placeholder authors.  Will it be confusing that those are existing GitHub accounts though?  `johndoe` appears to be a real user.

dhimmel:
> Will it be confusing that those are existing GitHub accounts though?

I considered that @johndoe and @janeroe were actual GitHub accounts. We could use a worse handle that's not taken, but then someone could always register it. Both @johndoe and @janeroe are inactive accounts it appears. Also you gotta expect this issue when you register those handles (;

dhimmel:
I'm going to rebase merge this. If people notice any additional issues, I'm happy to revise in a future PR.

agitter:
> Also you gotta expect this issue when you register those handles

Ha, yeah that's true.

PR #65, opened 2017-08-31
WIP: Squash all commits cloning method

dhimmel:
[DO NOT MERGE]

This is an experimental PR to investigate a "sqaush all" method for cloning this repo. When using this method, all manubot-rootstock commits up to the cloning moment are squashed. The result is like https://github.com/dhimmel/rephetio-manuscript/commit/4ae7a48b42446a9fc52ab0f71b0e5dce8af089fd.

The benefit of this method is that the commit history of cloned manuscripts is less cluttered.

Still looking into how merges will work... presumably we want to squash those as well.


rgieseke:
Is there an easy way to make an update for repos created this way? Just download the zip and overwrite files?

dhimmel:
> Is there an easy way to make an update for repos created this way?

I think I played around with this once, but didn't immediately find a perfect solution. However, I'm thinking the best solution would be to [cherry pick](https://git-scm.com/docs/git-cherry-pick) the desired range of commits (i.e. commits not yet included in your manuscript instance). After you finish the cherry picking, dealing with any issues, then perhaps you could squash all the newly added commits.

> Just download the zip and overwrite files?

This isn't ideal as it'll clobber any changes you've made to these files. Perhaps however, you could do this and then do `git add --patch` and select which part of the diff to commit and preserve. Afterwards, you could reset to get rid of the uncommitted changes.

rgieseke:
Thanks @dhimmel - I went with manually copying over and `git add --patch` :-)

PR #68, opened 2017-09-04
Customize Manubot citation style

dhimmel:
Simplify the CSL specification as per https://github.com/greenelab/manubot-rootstock/issues/53.

The two issues I can't resolve are:

1. Hyperlinking titles. Unfortunately, the [CSL docs](http://docs.citationstyles.org/en/stable/specification.html) don't mention the option to use a URL as a hyperlink. I'm guessing that hyperlinks are not a feature of CSL.
2. Making author names lighter or smaller.

They're not deal-breakers but would be nice.

You can see what this style currently looks [like here](https://github.com/dhimmel/rephetio-manuscript/blob/def8b4602260b38d27bb4b60d54c179fc03ee3cb/manuscript.pdf):

> bibliography

### DIFF
--- a/build/assets/style.csl
+++ b/build/assets/style.csl
@@ -1,142 +1,37 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-GB">
+<?xml version="1.0" encoding="utf-8"?>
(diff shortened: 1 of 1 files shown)


dhimmel:
@agitter do you have any objections to this change? Should we merge now and iterate / enhance the style over time?

agitter:
I have no objections.  Users are free to change the CSL file with whatever they prefer or is required by a journal, so this is mostly a matter of setting good defaults.  These look good to me.

Issue #73, opened 2017-09-28
MIT licensed JS code


rgieseke:
The `anchor` script (https://github.com/greenelab/manubot-rootstock/blob/master/build/assets/anchors.js) is not yet listed in the licensing section of the Readme.

I think it would be good to mention this somehow, maybe with "unless noted explicitly" or something similar?

dhimmel:
Good point & I agree.

Want to make a PR? If not, I'll do it.

rgieseke:
Done in #75

PR #75, opened 2017-09-28
Note MIT licensed 3rd-party code

rgieseke:
#73

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ We encourage users to openly license their manuscripts, which is the default as 
 [![License: CC BY 4.0](https://img.shields.io/badge/License%20All-CC%20BY%204.0-lightgrey.svg)](http://creativecommons.org/licenses/by/4.0/)
 [![License: CC0 1.0](https://img.shields.io/badge/License%20Parts-CC0%201.0-lightgrey.svg)](https://creativecommons.org/publicdomain/zero/1.0/)
 
-This entirety of this repository is licensed under a CC BY 4.0 License ([`LICENSE.md`](LICENSE.md)), which allows reuse with attribution.
+Except when noted otherwise the entirety of this repository is licensed under a CC BY 4.0 License ([`LICENSE.md`](LICENSE.md)), which allows reuse with attribution.
 Please attribute by linking to https://github.com/greenelab/manubot-rootstock.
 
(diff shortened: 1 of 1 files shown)


rgieseke:
Here's a proposal, happy to revise.

README.md
```diff
@@ -82,7 +82,7 @@ We encourage users to openly license their manuscripts, which is the default as
 [![License: CC BY 4.0](https://img.shields.io/badge/License%20All-CC%20BY%204.0-lightgrey.svg)](http://creativecommons.org/licenses/by/4.0/)
 [![License: CC0 1.0](https://img.shields.io/badge/License%20Parts-CC0%201.0-lightgrey.svg)](https://creativecommons.org/publicdomain/zero/1.0/)
 
-This entirety of this repository is licensed under a CC BY 4.0 License ([`LICENSE.md`](LICENSE.md)), which allows reuse with attribution.
+Except when noted otherwise the entirety of this repository is licensed under a CC BY 4.0 License ([`LICENSE.md`](LICENSE.md)), which allows reuse with attribution.
```
dhimmel: Comma after otherwise.

README.md
```diff
@@ -103,4 +103,8 @@ All other files are only available under CC BY 4.0, including:
 + `*.pdf`
 + `*.docx`
 
+Third-party code under an MIT license:
```
dhimmel: Perhaps change this to

```md
Except for the following files with different licenses:

+ `build/assets/anchors.js` which is released under an [MIT License](https://opensource.org/licenses/MIT).
```

This way we can add additional files with their own distinct licenses. Does that make sense?

dhimmel:
Looks like build failed due to https://stackoverflow.com/a/46457813/4651668. Not related to the PR.

README.md
agitter: I like @dhimmel's extended version.  We could also link back to the original source in the readme even if the MIT license requirements are already satisfied in `anchors.js`.  For example, change `build/assets/anchors.js` to [`build/assets/anchors.js`](https://github.com/bryanbraun/anchorjs).

dhimmel: I don't like linking what seems to be a file path to a GitHub repo landing page. If you want to attribute the source here, which I don't is necessary but I'm not against, you could hyperlink the word released rather than the file path. We can link the file path, but it should be a relative link to `build/assets/anchors.js`, which will bring someone to the file.

rgieseke: Added with your suggestions, thanks! I've linked to the Anchorjs landing page though.

agitter:
Thanks @rgieseke!

PR #78, opened 2017-10-03
PDF export: use weasyprint not wkhtmltopdf

dhimmel:
Closes https://github.com/greenelab/manubot-rootstock/issues/77

### DIFF
--- a/build/environment.yml
+++ b/build/environment.yml
@@ -8,7 +8,9 @@ dependencies:
   - anaconda::python=3.6.2
   - anaconda::pyyaml=3.12
   - anaconda::requests=2.14.2
-  - bioconda::wkhtmltopdf=0.12.3
+  - conda-forge::cairo=1.14.6
+  - conda-forge::cairocffi=0.7.2
+  - conda-forge::cffi=1.10.0
   - conda-forge::pandoc=1.19.2
   - conda-forge::watchdog=0.8.3
   - pip:
@@ -24,3 +26,4 @@ dependencies:
     - pysha3==1.0.2
     - python-bitcoinlib==0.8.0
     - requests-cache==0.4.13
+    - weasyprint==0.40
--- a/build/environment.yml
+++ b/build/environment.yml
@@ -9,7 +9,6 @@ dependencies:
   - anaconda::pyyaml=3.12
   - anaconda::requests=2.14.2
   - conda-forge::cairo=1.14.6
-  - conda-forge::cairocffi=0.7.2
   - conda-forge::cffi=1.10.0
   - conda-forge::pandoc=1.19.2
   - conda-forge::watchdog=0.8.3
--- a/USAGE.md
+++ b/USAGE.md
@@ -66,6 +66,8 @@ For example, the following will override the figure number to be "S1" and set th
 {#fig:supplement tag="S1" width="5in"}
 ```
 
+We recommend always specifying the width of SVG images (even if just `width="100%"`), since otherwise SVGs may not render properly in the [WeasyPrint](http://weasyprint.org/) PDF export.
+
 ### Citations
 
 Manubot supports Pandoc [citations](http://pandoc.org/MANUAL.html#citations) via `pandoc-citeproc`.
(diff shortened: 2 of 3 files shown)


dhimmel:
Compare `weasyprint` to `wkhtmltopdf` on the Sci-Hub Manuscript:

+ [manuscript-weasyprint.pdf](https://github.com/greenelab/manubot-rootstock/files/1354074/manuscript-weasyprint.pdf)
+ [manuscript-wkhtmltopdf.pdf](https://github.com/greenelab/manubot-rootstock/files/1354073/manuscript-wkhtmltopdf.pdf)

As you can see on the right, WeasyPrint fixes the SVG issues and the weird text distortion of `wkhtmltopdf`:

weasy-compare

It turns out however, that specifying the width or height in image imbed is required for proper SVG sizing (see https://github.com/Kozea/WeasyPrint/issues/442). So while not perfect, the SVG issues are much less than before.

Finally the big plus of WeasyPrint is that the conda environment should no longer be linux only.

dhimmel:
Here's the Project Rephetio manuscript converted using weasyprint: [manuscript-weasyprint.pdf](https://github.com/greenelab/manubot-rootstock/files/1354096/manuscript-weasyprint.pdf). This manuscript uses PNG images for figures, which weasyprint does well with.

agitter:
I didn't debug this yet, but this is what I saw in Windows 10 when I tried creating the new environment:
```
conda env create --file environment.yml
Fetching package metadata .............
Solving package specifications: .

NoPackagesFoundError: Package missing in current win-64 channels:
  - cairocffi 0.7.2*
```

dhimmel:
@agitter that's good to know. The issues is that cairocffi has no windows [conda-forge build](https://anaconda.org/conda-forge/cairocffi). It looks like it [can be installed](https://cairocffi.readthedocs.io/en/latest/overview.html#installing-cairo-on-windows) on windows, but I'm not sure whether a `pip install cairocffi` will work.

Can you remove `cairocffi` from `environment.yml`, and see if the install succeeds? This worked on Ubuntu 17.04, so possible we could fix this by using pip for `cairocffi`.

agitter:
The pip install doesn't work for me because I don't have this Windows machine set up for development.  It doesn't have Visual C++.  I'm sure I could get it to work if I wanted, but I don't think we want to assume users would have the necessary Windows compilers.

Other [third party conda packages](https://anaconda.org/search?q=platform%3Awin-64+cairocffi) only support Python 2.7.  If I create a Python 2.7 environment then `conda install -c rmg cairocffi=0.7.2` works, but that doesn't help us.

dhimmel:
@agitter given that we may be far away from Windows compatibility (e.g. will our shell commands work on windows), I think we should proceed here. If someone wants to tackle windows compatibility in the future, than we'll see.

I didn't realize python on Windows doesn't always come with `pip`. Or are you saying `pip install cairocffi` requires Visual C++?

Anyways, does this PR look ready to go?

agitter:
I agree that we don't want to support Windows now.  It would have been a bonus if this worked, but let's not get side-tracked.

I should have been more specific about the failure.  I have pip on this Windows machine, but the pip install failed with
```
 line 493, in _find_latest_available_vc_ver
        raise distutils.errors.DistutilsPlatformError(err)
    distutils.errors.DistutilsPlatformError: Microsoft Visual C++ 14.0 is required. Get it with "Microsoft Visual C++ Build Tools": http://landinghub.visualstudio.com/visual-cpp-build-tools
```

build/environment.yml
```diff
@@ -8,7 +8,8 @@ dependencies:
   - anaconda::python=3.6.2
   - anaconda::pyyaml=3.12
   - anaconda::requests=2.14.2
-  - bioconda::wkhtmltopdf=0.12.3
+  - conda-forge::cairo=1.14.6
+  - conda-forge::cffi=1.10.0
```
agitter: Do you want to add the cairocffi install back now that we will stay Linux-specific?  The environment [doesn't require it to be specified explicitly](https://travis-ci.org/greenelab/manubot-rootstock/builds/283358546#L599) so this should be fine as is.

dhimmel: Given the lack of windows binaries, I'm fine with allowing pip to do its thing here. There's lot's of dependencies that we don't explicitly pin. I'm still unsure of what the right balance is here... conda doesn't really have documentation for these sorts of decisions (:

PR #79, opened 2017-10-05
Add in width to images to better size images and fix warping

vsmalladi:
Attempt to fix warping of the images in docx.


dhimmel:
See https://github.com/greenelab/manubot-rootstock/pull/40#issuecomment-317280194 for the problem this attempts to fix.

PR #81, opened 2017-10-31
Update environment and documentation

dhimmel:
Pandoc upgraded to v2.0.0.1.

Swap anaconda packages to conda-forge.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -120,7 +120,7 @@ For guidance on what CSL JSON should be like for different document types, refer
 ## Manuscript metadata
 
 [`content/metadata.yaml`](content/metadata.yaml) contains manuscript metadata that gets passed through to Pandoc, via a [`yaml_metadata_block`](http://pandoc.org/MANUAL.html#extension-yaml_metadata_block).
-`metadata.yaml` should contain the manuscript `title`, `authors` list, and `keywords`.
+`metadata.yaml` should contain the manuscript `title`, `authors` list, `keywords`, and language (`lang`).
 Additional metadata, such as `date`, will automatically be created by the Manubot.
 Manubot uses the [timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) specified in [`build.sh`](build/build.sh) for setting the manuscript's date.
 For example, setting the `TZ` environment variable to `Etc/UTC` directs the Manubot to use Coordinated Universal Time.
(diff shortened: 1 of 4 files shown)


USAGE.md
```diff
@@ -120,7 +120,7 @@ For guidance on what CSL JSON should be like for different document types, refer
 ## Manuscript metadata
 
 [`content/metadata.yaml`](content/metadata.yaml) contains manuscript metadata that gets passed through to Pandoc, via a [`yaml_metadata_block`](http://pandoc.org/MANUAL.html#extension-yaml_metadata_block).
-`metadata.yaml` should contain the manuscript `title`, `authors` list, and `keywords`.
+`metadata.yaml` should contain the manuscript `title`, `authors` list, `keywords`, and language (`lang`).
```
agitter: @dhimmel can we link to a list of valid language codes?

dhimmel:
conda-forge is an open system, whereas the anaconda recipes to build packages are not public ([see comment here](https://jakevdp.github.io/blog/2016/08/25/conda-myths-and-misconceptions/#Myth-#8:-But-Conda-packages-themselves-are-closed-source,-right?)). In the past, anaconda had greater package coverage, but lately I think conda-forge now has better coverage and generally is updated with new versions more quickly.

In addition, if conda-forge is outdated, you can update it yourself, like I did in https://github.com/conda-forge/pandoc-feedstock/pull/19. Previously, we were only using conda-forge for packages that weren't available in anaconda. However, once I experienced an error from using an R binary from the anaconda R channel and an R package from conda-forge. So I think it's ideal to rely on a single channel if possible. Now conda-forge seems to be mature enough and up to the task.

USAGE.md
dhimmel: From [Pandoc docs](http://pandoc.org/MANUAL.html#language-variables):

> `lang`: identifies the main language of the document, using a code according to [BCP 47](https://tools.ietf.org/html/bcp47) (e.g. `en` or `en-GB`). For some output formats, pandoc will convert it to an appropriate format stored in the additional variables `babel-lang`, `polyglossia-lang` (LaTeX) and `context-lang` (ConTeXt).

I will look for a BCP 47 list of language codes.

dhimmel: BCP 47 codes can combine a language and region. Language codes are [defined here](http://www.loc.gov/standards/iso639-2/php/code_list.php). Region codes [are here](https://en.wikipedia.org/wiki/ISO_3166-1).

These BCP 47 codes can get quite complex [it seems](https://www.w3.org/International/articles/language-tags/) as they can have the following parts:

> language-extlang-script-region-variant-extension-privateuse

There's an unofficial validator at http://schneegans.de/lv/ that may be helpful. Wonder how complicated we want to make these docs!

agitter:
> conda-forge is an open system..

Thanks, that's a very helpful explanation.  I now recall a blog post that described what you noted about mixing channels.  On a related note, the [Anaconda 5.0 release notes](https://www.anaconda.com/blog/developer-blog/announcing-the-release-of-anaconda-distribution-5-0/) state

>Where possible, all build recipes are now using conda-forge as a base, via https://github.com/AnacondaRecipes.

USAGE.md
agitter: Great info.  Linking to one or two of those resources could help users without adding much more complexity to these manubot docs.

dhimmel: Fix proposed in https://github.com/greenelab/manubot-rootstock/pull/83

Issue #74, opened 2017-09-28
Setup commands fail on macOS


dhimmel:
Some of the commands in [SETUP.md](https://github.com/greenelab/manubot-rootstock/blob/3a42f6bd45fad874b29898cc744caf0ed23437af/SETUP.md) fail on macOS. IIRC, these commands are:

```sh
TRAVIS_ENCRYPT_ID=`grep \
  --only-matching --perl-regexp \
  --regexp='(?<=encrypted_)[a-zA-Z0-9]+(?=_key)' \
  travis-encrypt-file.log`
sed --in-place "s/f2f00aaf6402/$TRAVIS_ENCRYPT_ID/g" deploy.sh
```

```sh
sed --in-place "s/greenelab/$OWNER/g" README.md
sed --in-place "s/manubot-rootstock/$REPO/g" README.md
```

The issue is likely that the mac versions of these utilities don't support the same long arguments. What a shame.

dhimmel:
We could force mac users to switch their default `sed` and `grep` binaries to the GNU versions:

```sh
brew install gnu-sed --with-default-names
brew install grep --with-default-names
```

vsmalladi:
Testing on a mac and having homebrew helps fix some of the missing libraries:
cairo
pango

dhimmel:
> Testing on a mac and having homebrew helps fix some of the missing libraries

Are you saying that the conda environment installed on macOS but even then there were missing libraries?

Also `cairo` and `pango` are not relevant at the setup stage. This is for configuring a new manubot instance... before the conda environment is even installed.

Although you do potentially raise an interesting point. We do the SETUP commands in the conda environment to get consistent versions of [`sed`](https://anaconda.org/conda-forge/sed) and [`grep`](https://anaconda.org/bioconda/grep) (linux and macOS). However, I'm not sure it's worth requiring conda for setup... right now its just needed to build.

agitter:
@dhimmel are you thinking we would include sed and grep in `environment.yml` or just recommend the conda versions for Mac users?  A simple solution would be to add a troubleshooting section at the bottom of `SETUP.md`.  We can provide a sample `sed` call so that Mac users can test if they have a version that supports the required long arguments.  If not, we can show them how to install the GNU versions from brew or conda.

> Or perhaps we could make a setup script that runs in a docker image?

[...]

PR #84, opened 2017-11-05
Fix setup on MAC OSX

vsmalladi:
Fixes issues #74 for MAC OSX


dhimmel:
@vsmalladi it would probably be best to just have a single command that works on both linux and macOS. Did you test those commands on Linux? If not, I can try them out.

vsmalladi:
@dhimmel I haven't yet. If you can verify they work on Linux, then I can update the file to only have one set of commands.

agitter:
I tested this in Linux and the Git Bash client for Windows.  In both cases `sed -i '' "s/greenelab/$OWNER/g" README.md` gave me the error `sed: can't read s/greenelab/agitter/g: No such file or directory`.  The command ` sed --in-place "s/greenelab/$OWNER/g" README.md` worked.

dhimmel:
It looks like it [will be hard](https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed) to find cross-platform compatible `sed` commands. Let me look into possible python one liners

SETUP.md
```diff
@@ -137,6 +147,16 @@ sed --in-place "s/manubot-rootstock/$REPO/g" README.md
 git rm content/02.delete-me.md
 ```
 
+For setup on MAC OSX using BSD sed use the following commands.
+
+```sh
+sed -i '' "s/greenelab/$OWNER/g" README.md
```
dhimmel: What if we abandoned `--inplace`/`-1` and just ran a subseqent `mv` command?

SETUP.md
```diff
@@ -104,6 +104,16 @@ TRAVIS_ENCRYPT_ID=`grep \
 sed --in-place "s/f2f00aaf6402/$TRAVIS_ENCRYPT_ID/g" deploy.sh
 ```
 
+For setup on MAC OSX using BSD sed use the following commands.
```
dhimmel: We could try a python command like:

```sh
python -c "import re; print(re.search('encrypted_([a-zA-Z0-9]+)_key', open('README.md').read())).group(0)"
```

Not tested BTW.

SETUP.md
vsmalladi: We could have the first sed command output to tmp file and the send one take tmp and output into README.md, then follow with rm for tmp file.

dhimmel: If you use `mv -f`, you can skip the `rm` command.

SETUP.md
vsmalladi: Test the python and it works. Updating branch with new commands.

SETUP.md
```diff
@@ -97,11 +97,11 @@ The following command automates this substitution.
 
 ```sh
 # Edit ci/deploy.sh with travis secure env variables generated by travis encrypt-file
-TRAVIS_ENCRYPT_ID=`grep \
-  --only-matching --perl-regexp \
-  --regexp='(?<=encrypted_)[a-zA-Z0-9]+(?=_key)' \
-  travis-encrypt-file.log`
-sed --in-place "s/f2f00aaf6402/$TRAVIS_ENCRYPT_ID/g" deploy.sh
+TRAVIS_ENCRYPT_ID=`python -c "import re; \
+    print(re.search('encrypted_([a-zA-Z0-9]+)_key', \
```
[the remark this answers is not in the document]
dhimmel: I wonder if readability would improve with the following method:

```python
import re;
text = open('travis-encrypt-file.log').read();
print(re.search('encrypted_([a-f0-9]+)_key', text).group(1))
```

SETUP.md
```diff
@@ -97,11 +97,10 @@ The following command automates this substitution.
...
 ```sh
 # Edit ci/deploy.sh with travis secure env variables generated by travis encrypt-file
-TRAVIS_ENCRYPT_ID=`grep \
-  --only-matching --perl-regexp \
-  --regexp='(?<=encrypted_)[a-zA-Z0-9]+(?=_key)' \
-  travis-encrypt-file.log`
-sed --in-place "s/f2f00aaf6402/$TRAVIS_ENCRYPT_ID/g" deploy.sh
+TRAVIS_ENCRYPT_ID=`python -c "import re; \
+                  text = open('travis-encrypt-file.log').read(); \
+                  print(re.search('encrypted_([a-f0-9]+)_key', text).group(1))"`
```
dhimmel: Our style for the shell commands is 2 space indent

vsmalladi: Okay will update accordingly

dhimmel:
Thanks @vsmalladi. I tested the commands locally on Linux (not on a fully new manubot instances, but in a simulation). They worked on Ubuntu 17.10

dhimmel:
For future testing, here is an example `ci/travis-encrypt-file.log`:

```
encrypting deploy.key for greenelab/manubot-rootstock
storing result as deploy.key.enc
storing secure env variables for decryption

openssl aes-256-cbc -K $encrypted_1b0a5c5e82ff_key -iv $encrypted_1b0a5c5e82ff_iv -in deploy.key.enc -out deploy.key -d

Make sure to add deploy.key.enc to the git repository.
Make sure not to add deploy.key to the git repository.
Commit all changes to your .travis.yml.
```

I believe this was created with the `travis` gem v1.8.8.

PR #89, opened 2017-11-13
Update SETUP.md based on OpenCon do-a-thon

dhimmel:
Work in progress.

Thanks @zambujo @RaoOfPhysics @broadwym @schliebs @Hillea and others who've stopped by our OpenCon Do-A-Thon table. See https://github.com/sparcopen/doathon/issues/21 for more info.

Attendees have begun setting up manuscripts. We've noticed some pain points and difficulties. This PR aims to improve the `SETUP.md` documentation in response to these issues.

Feel free to review this PR or make additional doc suggestions.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -3,6 +3,7 @@
 The process to create a new Manubot manuscript is a bit challenging, because it requires a few steps that are difficult to automate.
 However, you will only have to perform these steps once for each manuscript.
 These steps should be performed in a terminal, starting in the directory where you want the manuscript folder be created.
+Windows users can use [Git Bash](https://git-for-windows.github.io/) to run the commands.
 
 ## Configuration
 
@@ -72,7 +73,7 @@ ssh-keygen \
 echo https://github.com/$OWNER/$REPO/settings/keys
 ```
 
-Manually add `deploy.key.pub` (with write access) to GitHub under the repository's deploy key settings (the URL echoed above).
+Manually add the text of `deploy.key.pub` (with write access) to GitHub under the repository's deploy key settings (the URL echoed above).
 Give the key a descriptive title, such as "Travis CI Manubot".
 
 For the next step, you need the [Travis command line client](https://github.com/travis-ci/travis.rb) installed.
--- a/SETUP.md
+++ b/SETUP.md
@@ -140,7 +140,7 @@ git rm content/02.delete-me.md
 
 ## Finalize
 
-Run `git status` or `git diff --word-diff` to check that the following files have unstaged changes:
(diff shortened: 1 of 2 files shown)


dhimmel:
> That's what I use on Windows.

Windows doesn't come with Ruby or Python right? So we're requiring users to install at least 2 languages... eek!

agitter:
That's right.  During my setups, I abandoned Windows for the Ruby/Travis steps though.  I don't recall if it was because of this [Travis caveat](https://docs.travis-ci.com/user/encrypting-files/#Caveat) or if I just thought it would be easier.

dhimmel:
> 👍  on Git Bash. That's what I use on Windows.

@agitter we may need to rethink our windows support based on the `deploy.key` encryption nightmare in https://github.com/greenelab/manubot-rootstock/issues/91

agitter:
Do we officially support Windows now?  Our conda environment doesn't.  I assumed we recommend Linux for setup as well, so I'm okay making that official.

I think that reaching the broadest user base possible would require highly-automated setup via a GitHub app instead of Windows command line support.

dhimmel:
> Do we officially support Windows now?

No. Although I think I should edit this PR to say that Windows may work with the Linux Subsystem, and otherwise will not work.

> Our conda environment doesn't

I thought it was possible that the conda environment now works on windows? From https://github.com/greenelab/manubot-rootstock/pull/78, it seems like the environment may work if visual C++ is installed?

> I think that reaching the broadest user base possible would require highly-automated setup via a GitHub app instead of Windows command line support.

Yes, I would venture to say that at the moment, most of the users with sufficient technical experience to setup a manuscript, probably do not exclusively use windows?

agitter:
> Windows may work with the Linux Subsystem, and otherwise will not work.

Agreed.  We could mention Docker as an untested option or avoid it until we've tested it.  [My Docker test](https://github.com/greenelab/manubot-rootstock/issues/6#issuecomment-319089862) was in Linux but provides a relevant container.

> I thought it was possible that the conda environment now works on windows?

Yes, it's likely to work if there is a C++ compiler available.  I was referring to our build readme, where we still say `Currently, the environment requires Linux.`  I agree with that message until we've successfully built a manuscript on Windows.

> most of the users with sufficient technical experience to setup a manuscript, probably do not exclusively use windows

👍

dhimmel:
> I was referring to our build readme, where we still say Currently, the environment requires Linux.

Updated this in https://github.com/greenelab/manubot-rootstock/pull/89/commits/a3b9206f484ca61f566fcfc4cadc648e4ae0b64e.

@agitter can you re-review with the latest changes?

PR #93, opened 2017-11-16
Use .html not .js for inserted HTML snippets

dhimmel:
While `analytics.html` and `anchors.html` do insert javascript into the HTML manuscript page, they are themselves HTML files I believe.

Perhaps @dongbohu, you can confirm that `.html` is a more appropriate extension?


dongbohu:
Since both `analytics.html` and `anchors.html` are enclosed in `<script>...</script>` tags, it makes more sense to name them with `html` extension than with `js` extension.

PR #97, opened 2018-01-09
Maintenance updates on 2018-01-09

dhimmel:
Intended for rebase merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ branches:
   only:
     - master
 before_install:
-  - wget https://repo.continuum.io/miniconda/Miniconda3-4.3.30-Linux-x86_64.sh
+  - wget https://repo.continuum.io/miniconda/Miniconda3-4.3.31-Linux-x86_64.sh
     --output-document miniconda.sh
   - bash miniconda.sh -b -p $HOME/miniconda
   - export PATH="$HOME/miniconda/bin:$PATH"
--- a/USAGE.md
+++ b/USAGE.md
@@ -135,7 +135,9 @@ initials: DSH  # strongly suggested
 orcid: 0000-0002-3012-7446  # mandatory
 twitter: dhimmel  # optional
 email: daniel.himmelstein@gmail.com  # suggested
-affiliations: Department of Systems Pharmacology and Translational Therapeutics, University of Pennsylvania  # strongly suggested
+affiliations:  # as a list, strongly suggested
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -5,7 +5,7 @@ branches:
   only:
     - master
 before_install:
-  - wget https://repo.continuum.io/miniconda/Miniconda3-4.3.30-Linux-x86_64.sh
+  - wget https://repo.continuum.io/miniconda/Miniconda3-4.3.31-Linux-x86_64.sh
```
agitter: I'm unclear on how the Miniconda versioning relates to the conda versioning.  Does this mean that we're still using conda 4.3 for Travis because 4.4 isn't available in a Miniconda installer yet?  I see that `Miniconda3-4.3.31-Linux-x86_64.sh` is the most recent installer.

Can we update conda to 4.4 after installing Miniconda so that our Travis install matches the readme and the new `conda activate manubot` syntax?

content/metadata.yaml
```diff
@@ -13,12 +13,15 @@ author_info:
...
     funders: Grant XXXXXXXX
   -
     github: janeroe
     name: Jane Roe
     initials: JR
     orcid: XXXX-XXXX-XXXX-XXXX
     email: jane.roe@whatever.edu
-    affiliations: Department of Whatever, University of Something
+    affiliations:
+      - Department of Something, University of Whatever
```
agitter: Great, I was going to suggest you make this change and see that you already did.

USAGE.md
```diff
@@ -135,7 +135,9 @@ initials: DSH  # strongly suggested
 orcid: 0000-0002-3012-7446  # mandatory
 twitter: dhimmel  # optional
 email: daniel.himmelstein@gmail.com  # suggested
-affiliations: Department of Systems Pharmacology and Translational Therapeutics, University of Pennsylvania  # strongly suggested
+affiliations:  # as a list, strongly suggested
```
agitter: We may want to explain why the `affiliations` differs from the other key-value pairs in the paragraph above.  Something about how affiliations are provided one per line so that each author can have one or more of them.

.travis.yml
dhimmel: Yeah I think you're understanding is correct. It'd be nice to have travis use `conda activate manubot`. However, my preference is to do this in the future, when a Miniconda is released that contains conda 4.4.

PR #98, opened 2018-01-11
Enable PubMed ID citations

dhimmel:
`pmcid` citations are higher suggested priority than `pmid` citations, because I think it's likely that the NCBI Citation Exporter will return higher quality metadata with greater reliability than our E-Utilities to CSL functionality.

Upgrade Manubot to https://github.com/greenelab/manubot/commit/b53f487026954729ce80ca963f3838ebca85a144

Refs https://github.com/greenelab/manubot/pull/24

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -76,9 +76,10 @@ Therefore, citations must be of the following form: `@source:identifier`, where 
 When choosing which source to use for a citation, we recommend the following order:
 
 1. DOI (Digital Object Identifier), cite like `@doi:10.15363/thinklab.4`.
-2. PubMed ID, cite like `@pmid:26158728`.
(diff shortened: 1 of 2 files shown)


USAGE.md
```diff
@@ -87,6 +88,8 @@ Here is a sentence with several citations [@doi:10.15363/thinklab.4; @pmid:26158
 ```
 
 Note that multiple citations must be semicolon separated.
+Be careful not to cite the same study using identifiers from multiple sources.
```
dhimmel: As per https://github.com/greenelab/manubot/pull/24#discussion_r160776333 by @agitter:

> Is the behavior that each is treated as a separate reference? We may want to note that behavior in the manubot-rootstock usage instructions.

PR #99, opened 2018-01-14
environment: explicitly set pandoc-xnos version

dhimmel:
Pinning pandoc-eqnos, pandoc-fignos, and pandoc-tablenos, but allowing them to trigger the installation of the latest pandoc-xnos version (which is a dependency) caused issues. See https://github.com/greenelab/deep-review/pull/774#issuecomment-357520693.

Thanks @tomduck for helping us diagnose the issue.


tomduck:
I wrote up an [Issue](https://github.com/tomduck/pandoc-fignos/issues/46) at pandoc-fignos and included a provisional plan to prevent this from happening again.

PR #101, opened 2018-01-16
Versioned webpage using commit directories

dhimmel:
Work in progress

Closes https://github.com/greenelab/manubot-rootstock/issues/96


dhimmel:
Testing this in a manubot instance at https://github.com/dhimmel/code-96

dhimmel:
> version directories are missing an index.html file. This should probably be a symlink or renaming of manuscript.html

completed

> SETUP.md: do we want to purge versions pre-fork?

Yes. completed

> ghp-import does not follow symlinks, which we require for the site to work as intended

This PR now works (see https://github.com/dhimmel/code-96) even though ghp-import doesn't preserve symlinks. If ghp-import implements an option to preserve symlinks, the diffs on the `webpage` branch would become more simple.

@agitter ready for review

agitter:
> ready for review

Okay. I'll likely need a few days to get to this.

dhimmel:
Here's an example commit on what upgrading OTS stamps now looks like https://github.com/dhimmel/code-96/commit/8dae193309ede8d00e62131d50b5f51095ccfa83. Requires a bitcoin full node.

dhimmel:
I created a script https://gist.github.com/dhimmel/5cf98acc58f60ede9504422e7a0a9f41 that will take a legacy gh-pages and create version directories for past commits. This is not required, but allows you to pull out historical versions into their own directories.

ci/deploy.sh
```diff
@@ -34,6 +29,16 @@ ssh-add ci/deploy.key
 git remote set-branches --add origin gh-pages output
 git fetch origin gh-pages:gh-pages output:output
 
+# Configure versioned webpage
+python build/webpage.py \
+  --checkout=gh-pages \
+  --version=$TRAVIS_COMMIT
+
+# Generate OpenTimestamps
```
agitter: I want to make sure I understand the timestamping strategy.  We still timestamp in the same way without `--wait` because we want to CI deploy script to terminate quickly.  However, we now have the ability to manually update all of the OTS Timestamps whenever we want as you did in https://github.com/dhimmel/code-96/commit/8dae193309ede8d00e62131d50b5f51095ccfa83 without rewriting history.  After the upgrade, those timestamps are no longer dependent on the calendar service.

Is that correct?  Do we need to document the upgrade instructions somewhere?

README.md
```diff
@@ -53,6 +53,9 @@ sh build/build.sh
 # when a change is detected.
 sh build/autobuild.sh
 
+# Configure the webpage directory
```
agitter: This appears to be an optional step for a local build.  If `sh build/build.sh` is still the minimal command needed to build the PDF and HTML manuscripts, we may want to indicate that more explicitly in the script comments or preceding text.

webpage/README.md
```diff
@@ -9,8 +11,11 @@ This directory contains the following files, which are mostly ignored on the `ma
 + [`index.html`](index.html) is an HTML manuscript.
 + [`github-pandoc.css`](github-pandoc.css) sets the display style for `index.html`.
 + [`manuscript.pdf`](manuscript.pdf) is a PDF manuscript.
-+ `*.ots` files are OpenTimestamps which can be used to verify manuscript existence at or before a given time.
-  [OpenTimestamps](opentimestamps.org) uses the Bitcoin blockchain to attest to file hash existence.
+
+The [`v`](v) directory contains directories for each manuscript version.
```
agitter: Do we want `v` to be a link if it is ignored on master?

ci/deploy.sh
dhimmel: Yes your understanding is correct. `--wait` is not feasible for CI builds as it could take days IIRC. Will update documentation.

dhimmel:
Note the `ots upgrade` command added to the documentation in this PR may fail due to https://github.com/opentimestamps/opentimestamps-client/issues/71.

slochower:
I think there may be an issue with this commit. I'll be brief because I have to run, but can fill in details later if the fix is not obvious to you.

I'm running locally with `sh build/autobuild.sh` and viewing the local HTML output with `python -m http.server` in `webpage/`. If I update my `content` directory, I see the autobuild run correctly and can verify the word count increasing. However, the webpage I see in my browser is not updated. Upon looking closer, it does not seem that the `html` files in `webpage/v` are being updated correctly at all. For example, I just added the text "(try 2)" to one of my Markdown files, hit save, and saw the build complete. But if I look for that phrase in my `webpage/v` directory using `find . -name "*.html" -exec grep -Hni "(try 2)" {} \;`, I get nothing. Since PDF is built from the webpage, I don't see my changes there, either. I can't be sure which commit caused this, but I think it may be related to this feature.

I can dig deeper later, if you are unable to confirm this behavior. FWIW, here is the output of `autobuild.sh`:

```bash
Retrieving and processing reference metadata
## INFO
Manuscript content parts:
00.front-matter
01.recent-observations
99.back-matter
## INFO
0 unique citations strings extracted from text
0 unique standard citations    
## INFO
requests-cache starting with 0 cached responses
## INFO
requests-cache finished with 0 cached responses
## INFO
Using UTC timezone.
Dating manuscript with the current datetime: 2018-02-17T01:15:54.189811+00:00
## INFO
Reading user-provided templating variables complete:
{}
## INFO
Generated manscript stats:
{
  "reference_counts": {
    "total": 0
  },
  "word_count": 1361
}
Exporting HTML manuscript
Exporting PDF manuscript
WARNING: Ignored `-ms-text-size-adjust: 100%` at 78:5, unknown property.
WARNING: Ignored `-webkit-text-size-adjust: 100%` at 79:5, unknown property.
WARNING: Ignored `-moz-box-sizing: content-box` at 204:5, unknown property.
WARNING: Ignored `-webkit-appearance: button` at 379:5, unknown property.
WARNING: Ignored `cursor: pointer` at 380:5, the property does not apply for the print media.
WARNING: Ignored `cursor: default` at 389:5, the property does not apply for the print media.
WARNING: Ignored `-webkit-appearance: textfield` at 410:5, unknown property.
WARNING: Ignored `-moz-box-sizing: content-box` at 411:5, unknown property.
WARNING: Ignored `-webkit-box-sizing: content-box` at 412:5, unknown property.
WARNING: Ignored `-webkit-appearance: none` at 423:5, unknown property.
WARNING: Invalid or unsupported selector 'button::-moz-focus-inner,
input::-moz-focus-inner ', Unknown pseudo-element: -moz-focus-inner
WARNING: Invalid or unsupported selector '*:not("#mkdbuttons") ', (<FunctionBlock not( … )>, ':not() only accepts a simple selector')
WARNING: Ignored `-webkit-font-smoothing: subpixel-antialiased` at 486:5, unknown property.
WARNING: Ignored `-moz-border-radius: 3px` at 491:5, unknown property.
WARNING: Ignored `-webkit-border-radius: 3px
` at 492:5, unknown property.
WARNING: Ignored `-webkit-font-smoothing: subpixel-antialiased` at 528:5, unknown property.
WARNING: Ignored `cursor: text
` at 529:5, the property does not apply for the print media.
WARNING: Ignored `word-break: break-all` at 733:5, unknown property.
WARNING: Ignored `word-break: break-word` at 734:5, unknown property.
WARNING: Ignored `-webkit-hyphens: auto` at 735:5, unknown property.
WARNING: Ignored `-moz-hyphens: auto` at 736:5, unknown property.
Build complete

```

dhimmel:
Thanks for the report @slochower.

You need to run the following to touch the `webpage` directory:

```sh
python build/webpage.py
```

Previously, `index.html` was a symlink to the output directory. But now `index.html` requires `webpage.py` to be updated.

Sorry to break your autobuild workflow. How useful have you found the autobuild script? Since webpage.py accepts arguments, it's not entirely clear to me how we should fix it. Probably add `python build/webpage.py` with the default arguments.

slochower:
Thanks for the quick reply. Interesting change. I agree that we should probably put `python build/webpage.py` in `autobuild.sh` and I think the default arguments are fine (at least for me) because when I run `autobuild.sh` it's usually because I'm trying to do look at something quickly locally. 

I mostly use this local workflow when I want to see my rendered document (for example, if I have a bunch of equations and figures). This comes in handy for more informal cases than writing a collaborative manuscript. Sometimes I want to write a complicated email with figures and syntax highlighting and so I start one of these Markdown documents -- but it doesn't make sense to use an existing `manubot-rootstock` directory for a manuscript and I don't necessarily want to setup a new GitHub repository just for a single document, so I lazily clone `manubot-rootstock` and just run it in local mode to generate an HTML or PDF document that I can send.

PR #102, opened 2018-01-16
Update usage instructions for citation tags

agitter:
Per https://github.com/greenelab/deep-review/issues/699#issuecomment-358079022 I believe `=` requires a citation tag and would like to document this.


dhimmel:
Refs https://github.com/greenelab/manubot-rootstock/issues/2

PR #113, opened 2018-02-17
autobuild.sh: include webpage.py execution

dhimmel:
Most autobuild users will want to automatically update the webpage. See discussion at
https://github.com/greenelab/manubot-rootstock/pull/101#issuecomment-366405495

Update README to better document autobuild.

@slochower can you take a look and confirm that this does what you want it to?


slochower:
Perfect.

PR #114, opened 2018-02-26
front matter: hyperlink permalink for brevity

dhimmel:
The preambles were getting a bit long. For example, see https://greenelab.github.io/scihub-manuscript/v/8e538a98e4f16bab632c00efb1fb4f1b23a1c122/

> not-brevity

### DIFF
--- a/content/00.front-matter.md
+++ b/content/00.front-matter.md
@@ -10,14 +10,15 @@ _A DOI-citable version of this manuscript is available at <https://doi.org/DOI_H
 
 {## Template to insert build date and source ##}
 <small><em>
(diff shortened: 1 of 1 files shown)


dhimmel:
To test this locally, but produce the preamble we'd expect from CI deployment, I run the following first:

```sh
export TRAVIS=true
export TRAVIS_REPO_SLUG=greenelab/manubot-rootstock
export TRAVIS_COMMIT=954363ae1473ed67edff838670f30ec0884b3212
```

permalink

@agitter what do you think? Is there some other way of conveying the permalink that you think'll be best?

PR #119, opened 2018-03-27
travis login information added

adebali:


agitter:
Thanks for the pull request @adebali.  Do you know whether the login is only required under certain conditions?  It has been a few months since I last followed our setup instructions for a new manuscript.  For the last one I created, my notes don't show anything about the Travis login.

SETUP.md
```diff
@@ -88,6 +88,9 @@ For the next step, you need the [Travis command line client](https://github.com/
 This program is a Ruby gem:
 install it with `gem install travis` (not `apt install travis`, which is a different program).
 
+After the install, you will need to provide your credentials to login to travis with
+```travis login --org```
```
agitter: Can wrap in a single ` instead of the triple ```

dhimmel: Let's keep this in the same paragraph as the previous sentence. After switching to single backticks rather than tripple, you can also put that code on the same as the text.

Thanks for helping with doc!

adebali:
Travis login could be required for the first time setup only. I followed the `SETUP.md` instructions, everything worked great except for that. 

Thanks for this tool! I was looking for something like this. It looks very promising.

SETUP.md
agitter: @adebali I can merge this if you please make the small changes above.  Thanks again.

agitter:
@dhimmel are we okay to merge?  I didn't test the behavior.  I have had to do Travis setup in a Docker container in the past because it doesn't work well on my standard machines.

agitter:
Well that didn't work.  I made a squash merge commit with commit message `travis login information added (https://github.com/greenelab/manubot-rootstock/pull/119)` but the commit message on f4cc39a481f8811a53832ed00272f69d4e269615 showed up as `travis login information added (#119)`.  Did I need to escape the URL somehow?

Sorry about that @dhimmel.  I agree it is confusing to have relative links only when these commits are duplicated in other repos.

dhimmel:
@agitter you're good. The actual commit has the full URL which is what matters. See https://github.com/greenelab/manubot-rootstock/commit/f4cc39a481f8811a53832ed00272f69d4e269615.patch. It's just the GitHub web interface that shortens it.

agitter:
Oh, good!

PR #138, opened 2018-10-09
Case sensitive environment variables

### LINKED ISSUE #137: Point out that OWNER is case sensitive
`OWNER=benjamin-lee` resulted in 

```
Benjamins-MacBook-Pro-9:ci BenjaminLee$ travis encrypt-file   --repo=$OWNER/$REPO   --force   deploy.key > travis-encrypt-file.log
repository not known to https://api.travis-ci.org/: benjamin-lee/ten_simple_rules_for_ml_in_bio
```

whereas `travis encrypt-file   --repo=Benjamin-Lee/$REPO   --force   deploy.key > travis-encrypt-file.log` worked.

agitter:
Closes #137 

Based on some quick testing with the URLs, I believe GitHub is not case-sensitive but Travis is.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -9,6 +9,7 @@ Setup is supported on Linux and macOS, but [**not on Windows**](https://github.c
 
 First, you must configure two environment variables (`OWNER` and `REPO`).
 These variables specify the GitHub repository for the manuscript (i.e. `https://github.com/OWNER/REPO`).
+Both variables are case-sensitive in some of the setup steps.
 **Edit the following commands with your manuscript's information:**
 
 ```sh
--- a/SETUP.md
+++ b/SETUP.md
@@ -9,12 +9,12 @@ Setup is supported on Linux and macOS, but [**not on Windows**](https://github.c
 
 First, you must configure two environment variables (`OWNER` and `REPO`).
 These variables specify the GitHub repository for the manuscript (i.e. `https://github.com/OWNER/REPO`).
-Make sure that the case of `OWNER` matches how your account name is displayed on GitHub.
+Make sure that the case of `OWNER` matches how your username is displayed on GitHub.
 In general, assume that all commands in this setup are case-sensitive.
 **Edit the following commands with your manuscript's information:**
 
 ```sh
-# GitHub account (change from greenelab)
+# GitHub username (change from greenelab)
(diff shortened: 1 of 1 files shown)


SETUP.md
```diff
@@ -9,6 +9,7 @@ Setup is supported on Linux and macOS, but [**not on Windows**](https://github.c
 
 First, you must configure two environment variables (`OWNER` and `REPO`).
 These variables specify the GitHub repository for the manuscript (i.e. `https://github.com/OWNER/REPO`).
+Both variables are case-sensitive in some of the setup steps.
```
dhimmel: I wonder how helpful this is compared to saying:

1. things generally are case sensitive in this script
2. make sure to use your username capitalization from GitHub.

agitter: I could do that instead.  I believe the repo is also case sensitive.  Should I keep the comment here or move it earlier in the guide?

dhimmel: If I go to https://github.com/Greenelab/Manubot-rootstock, GitHub does not redirect to the canoical capitalization. However, if you click on any GitHub links, it contains the all lowercase capitalization we set. I wonder what the best way to describe this... "use the capitalization shown on GitHub"

At this point however, REPO does not exist... it gets created in the guide.

What about:

> Make sure to use the same casing for your username as gets displayed on GitHub. In general, assume that all commands in this setup are case sensitive.

agitter: Oh yes, I see.  I was testing Travis and GitHub repo case-sensitivity on existing repos.  Whatever case is provided here will be valid, but the user will then be stuck with it.

I revised your suggested phrasing in 5e2f179.

SETUP.md
```diff
@@ -9,6 +9,8 @@ Setup is supported on Linux and macOS, but [**not on Windows**](https://github.c
 
 First, you must configure two environment variables (`OWNER` and `REPO`).
 These variables specify the GitHub repository for the manuscript (i.e. `https://github.com/OWNER/REPO`).
+Make sure that the case of `OWNER` matches how your account name is displayed on GitHub.
```
dhimmel: Should we say username rather than "account name" which is how GitHub docs say it

agitter: I used account name to match the lines below.  I'll change it in both places.

PR #142, opened 2018-11-02
Ignore OS specific files

dhimmel:


dhimmel:
See for example https://github.com/trang1618/tpot-ds-ms/commit/a12a60b32db9a9263f4320465a2f5ace7dd2407c... if a user stages all files, it's easy to commit system files.

See also https://github.com/trang1618/tpot-ds-ms/commit/97c34e55180adbfde2975242c718a674967ddff4. Potentially we should ignore the `images` directory in the root.

dhimmel:
> Can you please remind me where the images directory in the root comes from. Is that created during a local build?

See line 47 and 63:

https://github.com/greenelab/manubot-rootstock/blob/895924e6a8c693a0ee110ec410b153b7ceb09bc2/build/build.sh#L45-L63

`images` is created in the root directory so the PDF pandoc call can locate the images. This has been a big annoyance. If the PDF generation fails, then `images` never gets removed. This causes the subsequent run of `ln -s` to fail. We probably got to fix this... so perhaps leave that out of .gitignore for now and aim for a more comprehensive solution.

agitter:
Thanks for the refresher.  I agree that may be out of scope for a `.gitignore` pull request but should be addressed.

PR #145, opened 2018-11-08
Support BUILD_PDF=false environment variable

dhimmel:
Closes https://github.com/greenelab/manubot-rootstock/issues/122 by @rgieseke.
Relates to https://github.com/greenelab/manubot-rootstock/issues/144#issuecomment-437119645 by @trang1618.

Several users have had issues building the PDF locally, especially on macOS. This PR enables a way to disable PDF builds, since users may be OKAY with just HTML output locally. `webpage.py` has been updated so as not to fail if `manuscript.pdf` is missing.

@rgieseke a review from you would be a appreciated since I think you're familiar with this part of the code?

TODO: investigate whether OTS timestamping on travis deploy would fail without a PDF.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -42,31 +42,34 @@ pandoc --verbose \
   --output=output/manuscript.html \
   $INPUT_PATH
 
-# Create PDF output
-echo "Exporting PDF manuscript"
-ln -s content/images images
-pandoc \
-  --from=markdown \
-  --to=html5 \
-  --pdf-engine=weasyprint \
-  --pdf-engine-opt=--presentational-hints \
-  --filter=pandoc-fignos \
-  --filter=pandoc-eqnos \
-  --filter=pandoc-tablenos \
-  --bibliography=$BIBLIOGRAPHY_PATH \
-  --csl=$CSL_PATH \
-  --metadata link-citations=true \
-  --webtex=https://latex.codecogs.com/svg.latex? \
-  --css=webpage/github-pandoc.css \
-  --output=output/manuscript.pdf \
-  $INPUT_PATH
-rm -r images
+# Create PDF output (unless BUILD_DOCX environment variable equals "false")
+if [ "$BUILD_PDF" != "false" ]
+then
+  echo "Exporting PDF manuscript"
+  ln -s content/images images
+  pandoc \
+    --from=markdown \
+    --to=html5 \
+    --pdf-engine=weasyprint \
+    --pdf-engine-opt=--presentational-hints \
+    --filter=pandoc-fignos \
+    --filter=pandoc-eqnos \
+    --filter=pandoc-tablenos \
+    --bibliography=$BIBLIOGRAPHY_PATH \
+    --csl=$CSL_PATH \
+    --metadata link-citations=true \
+    --webtex=https://latex.codecogs.com/svg.latex? \
+    --css=webpage/github-pandoc.css \
+    --output=output/manuscript.pdf \
+    $INPUT_PATH
+  rm -r images
+fi
 
-# Create DOCX output when user specifies to do so
+# Create DOCX output (if BUILD_DOCX environment variable equals "true")
 if [ "$BUILD_DOCX" = "true" ];
 then
-    echo "Exporting Word Docx manuscript"
-    pandoc --verbose \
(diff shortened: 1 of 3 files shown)


build/build.sh
```diff
@@ -42,31 +42,32 @@ pandoc --verbose \
...
-  --filter=pandoc-tablenos \
-  --bibliography=$BIBLIOGRAPHY_PATH \
-  --csl=$CSL_PATH \
-  --metadata link-citations=true \
-  --webtex=https://latex.codecogs.com/svg.latex? \
-  --css=webpage/github-pandoc.css \
-  --output=output/manuscript.pdf \
-  $INPUT_PATH
-rm -r images
+# Create PDF output (unless BUILD_DOCX environment variable equals "false")
```
agitter:
```suggestion
# Create PDF output (unless BUILD_PDF environment variable equals "false")
```

build/build.sh
```diff
@@ -42,31 +42,32 @@ pandoc --verbose \
...
-  --metadata link-citations=true \
-  --webtex=https://latex.codecogs.com/svg.latex? \
-  --css=webpage/github-pandoc.css \
-  --output=output/manuscript.pdf \
-  $INPUT_PATH
-rm -r images
+# Create PDF output (unless BUILD_PDF environment variable equals "false")
+if [ "$BUILD_PDF" != "false" ]; then
+  echo "Exporting PDF manuscript"
+  ln -s content/images images
```
dhimmel: @agitter this could also be a good PR to fix the image symlink issues.

So I think the main issue we've had is when the pandoc command fails, so `rm -r images` never occurs. This can lead to a leftover images symlink that prevents causes subsequent `ln -s content/images images` commands to fail.

One option would be to add the `-f` option to `ln`, causing it to overwrite an existing `images` symlink. However, there may be a risk if users somehow try running these commands in a different directory, this could delete their `content/images` directory and contents perhaps.

agitter: That option could work.  Many users may already have the `content/images` in version control so accidental deletion wouldn't be catastrophic.  But for anyone who didn't have them versioned and accidentally removed content with the script it would be a major problem.

I haven't come up with a better option though.  Perhaps the script could test whether `images` already exists and if it does, test whether it is a symbolic link.  There should be minimal harm in overwriting or removing a symbolic link.

dhimmel:
I have tested the following at https://github.com/greenelab/manubot-rootstock/commit/96270cd5bb63daf94e1e711af267ca3a0eadcf7d:

+ `sh build/build.sh` builds both the HTML & PDF
+ `BUILD_PDF=false sh build/build.sh` builds just HTML
+ killing `sh build/build.sh` during PDF generation, such that images symlink is not removed, and then running `sh build/build.sh` succeeds.
+ `touch images` and then running `sh build/build.sh` produces `ln: failed to create symbolic link 'images': File exists`

However, `mkdir images` followed by `sh build/build.sh` is misbehaved and gives error `rm: cannot remove 'images': Is a directory`. This is because the symlink gets created at `images/images`. Will look into `ln` docs.

dhimmel:
testing as of https://github.com/greenelab/manubot-rootstock/pull/145/commits/4208b1c2377d3a8fb5037cec5ea284219c7103db on Linux:

+ `mkdir images` and then `sh build/build.sh` produces `ln: failed to create symbolic link './images': File exists`
+ `touch images` and then  `sh build/build.sh` produces `ln: failed to create symbolic link './images': File exists`
+ `ln -s content/images` and then `sh build/build.sh` succeeds.

Let me track down a macOS to confirm the symlink behavior is the same.

dhimmel:
Can someone run the following two commands in a macOS terminal (without a custom shell) and report their output:

```
rm -rf test && touch test && ln -s example/test
rm -rf test && mkdir test && ln -s example/test
```

If all is good, they should both return errors.

trang1618:
Here's what I got:

```
$ rm -rf test && touch test && ln -s example/test
ln: ./test: File exists
$ rm -rf test && mkdir test && ln -s example/test
ln: ./test: File exists
```

dhimmel:
Nice, based on the errors on macOS in @trang1618's comment, I think the code is now GOOD TO GO

PR #148, opened 2018-11-16
SETUP.py: add checkout command to rootstock upgrade

dhimmel:
Addresses workflow error in https://github.com/greenelab/deep-review/pull/939#issuecomment-439220820


SETUP.md
```diff
@@ -188,3 +196,6 @@ This is helpful to see each upstream change.
 You may notice changes that effect how items in `content` are processed.
```
agitter:
```suggestion
You may notice changes that affect how items in `content` are processed.
```

Noticed this typo while inspecting the new instructions.

PR #150, opened 2018-12-10
webpage utility: attempt to upgrade timestamps

dhimmel:
I was mistaken in the past when I though upgrading a timestamp required a local Bitcoin node. That is not the case, only verifying a timestamp requires a node. Upgrading refers to adding information to the `.ots` file that connects the merkle tree root returned by the calendar to a bitcoin block header.

I had put off making this change due to https://github.com/opentimestamps/opentimestamps-client/issues/71, but I decided we should just do our own workaround, since the timestamps have a much larger trust-model until upgraded.

### DIFF
--- a/build/webpage.py
+++ b/build/webpage.py
@@ -22,12 +22,13 @@ def parse_arguments():
              "Generally a commit hash, tag, or 'local'. "
              "(default: '%(default)s')"
     )
-    parser.add_argument(
+    cache_group = parser.add_mutually_exclusive_group()
+    cache_group.add_argument(
         '--no-ots-cache',
         action='store_true',
         help="disable the timestamp cache."
     )
-    parser.add_argument(
+    cache_group.add_argument(
         '--ots-cache',
         default=pathlib.Path('ci/cache/ots'),
         type=pathlib.Path,
--- a/build/webpage.py
+++ b/build/webpage.py
@@ -18,7 +18,20 @@ def parse_arguments():
     parser.add_argument(
         '--version',
         default=os.environ.get('TRAVIS_COMMIT', 'local'),
-        help="Used to create webpage/v/{version} directory. Generally a commit hash, tag, or 'local'",
+        help="Used to create webpage/v/{version} directory. "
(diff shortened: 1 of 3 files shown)


dhimmel:
> If the cache is working, this does not affect build times, right?

Hopefully not much. The cache consists of many small files in many nested directories under `ci/cache/ots`. Travis is already caching the `ci/cache` directory, which also contains a database of cached requests. If the ots cache seems to be increasing continuous deploy times, we could disable it.

I may consider disabling it now in `deploy.sh`, because I don't think it should really be needed there, since I don't think it's needed for already upgraded timestamps.

dhimmel:
From [opentimestamps-client readme](https://github.com/opentimestamps/opentimestamps-client/blob/1b328269ceee66916e9a639e8d5d7d13cd70d5d8/README.md):

> The client maintains a cache of timestamps it obtains from remote calendars, so if you verify the same file again it'll use the cache:

From [`cmds.py`](https://github.com/opentimestamps/opentimestamps-client/blob/b51a9492c4a35f86f4c02371a0825e24087d371c/otsclient/cmds.py#L244-L247)

```
# First, check the cache for upgrades to this timestamp. Since the cache is
# local, we do this very agressively, checking every single sub-timestamp
# against the cache.
```

I don't think Travis needs to use a cache, because we upgrade timestamps such that they no longer need to query remote calendars. I don't see Travis running `ots upgrade` on the same incomplete timestamp multiple times. So I will disable caching on Travis, such that our Travis cache is not affected

dhimmel:
# Success

Merged into https://github.com/greenelab/manubot-rootstock/commit/aa664b4fa8974ef2f57f73593d781173b24243ea.

The Travis CI deployment upgraded timestamps, as seen in the [build log](https://travis-ci.org/greenelab/manubot-rootstock/builds/466585079#L1971).

The resulting gh-pages commit https://github.com/greenelab/manubot-rootstock/commit/98de8cfbeb106f216b028813112ee023e9bdbcf9 includes the upgraded timestamps.

Issue #156, opened 2019-01-10
Separate original pandoc CSS from manubot modified CSS?


vincerubinetti:
I'm thinking it may be helpful to separate the two somehow, either by putting our (manubot) CSS overrides at the bottom of `github-pandoc.css`, or even just in a separate css file. Or maybe in the html file insert I am writing to generate the tooltips, which also contains its own CSS for the tooltips.

Are there any objections to this? Also does someone have a quick summary of the CSS we "slowly modified over time", or would I have to go through the commits.

slochower:
No objection, but I'm not sure `github-pandoc.css` is really "pandoc CSS" in any sense. I think it's really just a CSS file that we pass to `pandoc` to make the rendered HTML look GitHub-like.

vincerubinetti:
Yes, sorry, I didn't notice that that stylesheet wasn't authored by pandoc themselves. Regardless, it might be wise to separate our modifications from any third party dependencies. Makes it more clear which party modified which thing, and makes it easier to update the dependencies separately without breaking/forgetting anything.

dhimmel:
The gist we copied github-pandoc.css from hasn't been updated in years. It was just a file someone posted and doesn't seem to be maintained.

Perhaps we could compare the file to its original version when added to this repo (ignoring whitespace). There haven't been too many changes.

Do you think github-pandoc.css is good or it's something we may want to rewrite from scratch.

vincerubinetti:
I'll review the css further today, but it seems relatively fine to me. Nothing jumps out at me as bad/dated, in terms of "code quality". 

That being said, it doesn't really look exactly like Github, if that's what we're going for. (Maybe it did look like it at the time it was written?). Having 2-3 themes of the most popular/common options, like Github, that PubCSS, etc per [#Issue 112](https://github.com/greenelab/manubot-rootstock/issues/112) might not actually be a bad idea. I could probably knock them out in 1-2 days per theme.

dhimmel:
As far as I'm aware, we are not interested in making our manuscripts adhere to the GitHub theme. It seems like the [original gist](https://gist.github.com/dashed/6714393) does have this objective. However, I think we just care about making a nice interface to read and interact with articles. When I named the file `github-pandoc.css`, I think I meant `github-pages-pandoc.css`. We could consider renaming this file to anything, if that would help remove confusion over its scope and goals.

vincerubinetti:
Closed due to complete rewrite of CSS theme.

[...]

PR #157, opened 2019-01-11
Display improvements

vincerubinetti:
Branch that contains my display (front-end) improvements to pandoc html output. Don't merge until we've developed a more comprehensive list of UI enhancements and completed them (added them to this branch)?

So far I'm putting all of these modifications in a single file called `manubot.html`, which contains both javascript (eg for generating/position tooltips) and css (for `github-pandoc.css` style overrides, tooltip styles, etc).


slochower:
When you're ready, it would be nice if you could add a screenshot to the PR highlighting your improvements.

vincerubinetti:
**Tooltips**
screenshot from 2019-01-11 15-18-41
Currently just a carbon copy of the ref at the bottom of the doc, with the same stylings as the rest of the document. Positioning defaults to above and to the right of the citation link. Switches horizontal/vertical sides if tooltip wont fit in viewport.

**Table scroll**
screenshot from 2019-01-11 15-22-01

Other "under-the-hood" changes contained in the commits. More display enhancements will be added to this PR before it is merged, unless @dhimmel wants to split them all up into smaller chunks (branches).

build/build.sh
```diff
@@ -39,6 +39,7 @@ pandoc --verbose \
   --include-in-header=build/assets/analytics.html \
   --include-after-body=build/assets/anchors.html \
   --include-after-body=build/assets/hypothesis.html \
+  --include-after-body=build/assets/manubot.html \
```
dhimmel: Is the addition of manubot.html to split the CSS that comes from the original gist and the manubot specific changes (as per #156)?

That is fine by me, but what do we do when there are mistakes in `github-pandoc.css`. It seems like modifying this file is a bit of an inevitability.

I also see this change will result in style information being directly inserted into the HTML output, rather than as a standalone `.css` file. There are some pros and cons to this? One pro is that we switch all CSS to being including in HTML output and get rid of having to manage `github-pandoc.css` as separate files on the gh-pages branch. However, this may increase the size of the git repo, since `github-pandoc.css` blobs only need to be stored once.

vincerubinetti:
Splitting up PR into multiple PR's per discussions with @dhimmel.

PR #160, opened 2019-01-18
webpage: add manubot.html with citation tooltips & scrolling tables

vincerubinetti:
- added tooltips
51058712-a8ff1a80-15b7-11e9-914b-d16c8adb2414

- added scrollbars to tables
51058830-0e530b80-15b8-11e9-8854-49799224bdf9

### DIFF
--- a/build/assets/manubot.html
+++ b/build/assets/manubot.html
@@ -1,0 +1,166 @@
+<script>
+
+	// insert dom node before another node
+    function insertAfter(newNode, referenceNode) {
+        referenceNode.parentNode.insertBefore(
+            newNode,
+            referenceNode.nextSibling
+        );
+    }
+    // remove all children of an element
+    function removeChildren(element) {
+        while (element.lastChild) {
+            element.lastChild.remove();
+        }
+    }
+    // get position of element relative to viewport
+    function getPositionInView(element) {
+        return {
+            left: element.getBoundingClientRect().left,
+            top: element.getBoundingClientRect().top,
+            right: element.getBoundingClientRect().right,
+            bottom: element.getBoundingClientRect().bottom
+        };
+    }
+    // get position of element relative to page
+    function getPositionInPage(element) {
+        return {
+            left:
+                element.getBoundingClientRect().left -
+                document.body.getBoundingClientRect().left,
+            top:
+                element.getBoundingClientRect().top -
+                document.body.getBoundingClientRect().top,
+            right:
+                element.getBoundingClientRect().right -
+                document.body.getBoundingClientRect().left,
+            bottom:
+                element.getBoundingClientRect().bottom -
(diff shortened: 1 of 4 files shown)


build/assets/manubot.html
```diff
@@ -0,0 +1,189 @@
+<script>
+
+	// insert dom node before another node
```
dhimmel: is comment misaligned?

vincerubinetti:
@dhimmel I was thinking it would make sense to bundle the `css` needed for each "plugin" with the plugin `javascript` itself. This would sort of decouple all the different features from each other. Here's an off-the-cuff list of what could be our plugins/themes:

- `default.css` - css only. default nice styling of pandoc html output. squashes tables to fit page width.
- `nih.css` - NIH grant style theme, per Casey's suggestion
- `academic.css` - typical "academic" paper style. serif font, two columns when printed
- `lens.css` - lens viewer looking theme. serif font, little more fancy/elegant looking.
- `table-scroll.html` - js and css. allows tables that are too wide for page width to scroll instead of being squashed
- `tooltips.html` - js and css. provides tooltips when hovering citation links
- `toc.html` - js and css. some kind of menu that will provide table of contents with links to jump to sections
- `samerefs.html` - js and css. some kind of menu that will show other instances of the selected reference. might be combined with `toc.html` menu?
- `accordion.html` - js and css. will allow `h2` sections to be collapsed/expanded
- `anchors.html` - js and css. puts permalinks next to desired heading sections.
- `hypothesis.html` - js and css. adds side bar for annotations.
- `analytics.js` - js. adds google analytics or similar tracking

This way all these things are separate, and are also somewhat clear in their function just by virtue of their file names. Users can toggle these things on/off just by including/removing them in/from the `build.sh` command (or command line argument, or manubot config file, if we ever switch to either of those approaches).

dhimmel:
@vincerubinetti this modular design makes sense to me. It makes enabling/disabling certain features easy. Hopefully there aren't too many interactions between various plugins that complicate things. It also gives us a nice way to divy up pull requests... we can start with `default.css`, then do `tooltips.html`, and so on.

Let's hold off on additional themes as I'm not sure there is much benefit to them yet. Would the different themes replace `default.css`? For example, `lens.css` would be used instead of `default.html`?

vincerubinetti:
If we design the plugins right, there shouldn't be any interactions in terms of functionality. However, I'm thinking now that the css for each plugin will probably have to be built into each theme instead of being bundled with the plugin itself. So for example, each theme file would look like this:
```css
<style>

/* styles for pure html */
h1, h2, h3, h4, h5, h6 {
    margin: 20px 0;
}

/* ... etc ... */

/* styles for manubot plugins */
.permalink { styles }
.tooltip { styles }
.toc { styles }

/* ... etc ... */

</style>
```

PR #162, opened 2019-01-22
add html5 section wrapper tag

vincerubinetti:
Adds command line argument top pandoc to wrap manuscript sections in the document (eg. abstract, authors, introduction, references, etc.) with the html5 `<section>` tag. This aligns with modern best practices of formatting html documents, and also allows us a bit more flexibility for styling/formatting. This shouldn't affect anything with our current CSS or javascript.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -60,6 +60,7 @@ if [ "$BUILD_PDF" != "false" ]; then
     --csl=$CSL_PATH \
     --metadata link-citations=true \
     --webtex=https://latex.codecogs.com/svg.latex? \
+    --section-divs \
     --css=webpage/github-pandoc.css \
     --output=output/manuscript.pdf \
     $INPUT_PATH
--- a/build/build.sh
+++ b/build/build.sh
@@ -35,6 +35,7 @@ pandoc --verbose \
   --csl=$CSL_PATH \
   --metadata link-citations=true \
   --mathjax \
+  --section-divs \
   --css=github-pandoc.css \
   --include-in-header=build/assets/analytics.html \
   --include-after-body=build/assets/anchors.html \


vincerubinetti:
After thinking about this more, I don't think there's a good solution.

Both the `<section>` and child `<h2>` probably _should_ have assigned ids, meaning that one will have to have some modifier like `-1` (note that adding "generational" dashes like this seems to be standard practice in assigning ids). We could remove the id from the section tag completely, which would involve either modifying the pandoc source somehow (pandoc is the one assigning these ids I believe) or modifying the pandoc output (removing the id from the `<section>` and re-formatting all the child ids). The former solution is ugly because it's modifying pandoc, and the latter solution is complicated and ugly because I'd have to check all the children and reassign numbers just in case there are subsections with the same name (resulting in "abstract-2" for example).

So, only two solutions are really: 1) go without sections and deal with the limitations that entails 2) use sections and deal with the slight ugliness of the dashes.

dhimmel:
> use sections and deal with the slight ugliness of the dashes

It sounds like sections will be valuable. What about modifying anchorsjs to remove a single `-1` if present at the end of a header? Or is that too hacky?

vincerubinetti:
> > use sections and deal with the slight ugliness of the dashes
> 
> It sounds like sections will be valuable. What about modifying anchorsjs to remove a single `-1` if present at the end of a header? Or is that too hacky?

That would create two elements with the same id. Ids in html are supposed to be unique identifiers, and you can get unexpected/glitchy behavior if you duplicate them.

dhimmel:
> That would create two elements with the same id

If I understand correctly, pandoc is creating all the ids. For example, with `--section-divs`, the following HTML will be created by pandoc:

```html
<section id="abstract">
<h2 id="abstract-1" class="page_break_before">Abstract</h2>
```

So my thinking was that we could modify anchorsjs to link to the id without the `-1` suffix, but not actually modify any ids.

Alternatively, could we use the `data-anchor-id` [feature of anchorsjs](https://www.bryanbraun.com/anchorjs/#section-ids). I guess that would require us to use javascript to add the right `data-anchor-id` to headers?

vincerubinetti:
> So my thinking was that we could modify anchorsjs to link to the id without the `-1` suffix, but not actually modify any ids.
> 
> Alternatively, could we use the `data-anchor-id` [feature of anchorsjs](https://www.bryanbraun.com/anchorjs/#section-ids). I guess that would require us to use javascript to add the right `data-anchor-id` to headers?

Oh I get what you mean now. I'd be wary of just removing the `-1`. What if a user had a heading like `Mapping of Auto-immune Virus HV-123-1`. I'll look into the `data-anchor-id`. That might allow a solution.

vincerubinetti:
Closed due to our new file structure of `/plugins` and `/themes`, where each added feature (like tooltips or table scrolling) will be its own `.html` file and PR.

PR #166, opened 2019-01-25
updated icons to font awesome and formatted more consistently

vincerubinetti:
Twitter and Github icons were slightly off center. ORCID icon formatting was a mess. All icons had multiple unnecessary properties.


content/images/github.svg
```diff
@@ -1,6 +1,3 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!-- Source: https://github.com/simple-icons/simple-icons/blob/95e3588af989ecf62f5f970af8fc7c64103ecd99/icons/github.svg -->
```
agitter: Do we want to leave an embedded comment with the new source?

vincerubinetti: Perhaps. In the `manubot-resources` repo I give attribution to Font Awesome, and say that all icons are by them unless otherwise noted. They don't include a url in their svgs by default; they keep it clean and to a bare minimum. Maybe we could include the url in the svgs that are not from FA.

dhimmel: My preference is to keep the attribution in the file.

vincerubinetti: Added the link back in. I also changed the fill color to `currentColor`. This will allow us to specify the color in the CSS instead of in the svg, and thus we can tweak the color for future themes without duplicating the svg files.

dhimmel: Let's add that change to this PR

dhimmel:
Is there an options to define these SVGs in one of our embedded HTML files? I wonder if that makes more sense then having them as separate files that clutter `images`? @vincerubinetti what's your thought on that?

content/images/github.svg
```diff
@@ -1,6 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!-- Source: https://github.com/simple-icons/simple-icons/blob/95e3588af989ecf62f5f970af8fc7c64103ecd99/icons/github.svg -->
-<svg aria-labelledby="title" role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" fill-rule="evenodd" clip-rule="evenodd" stroke-linejoin="round" stroke-miterlimit="1.414">
-  <title>GitHub icon</title>
-  <path fill="#8C8C8C" d="M12 .297c-6.63 0-12 5.373-12 12 0 5.303 3.438 9.8 8.205 11.385.6.113.82-.258.82-.577 0-.285-.01-1.04-.015-2.04-3.338.724-4.042-1.61-4.042-1.61C4.422 18.07 3.633 17.7 3.633 17.7c-1.087-.744.084-.729.084-.729 1.205.084 1.838 1.236 1.838 1.236 1.07 1.835 2.809 1.305 3.495.998.108-.776.417-1.305.76-1.605-2.665-.3-5.466-1.332-5.466-5.93 0-1.31.465-2.38 1.235-3.22-.135-.303-.54-1.523.105-3.176 0 0 1.005-.322 3.3 1.23.96-.267 1.98-.399 3-.405 1.02.006 2.04.138 3 .405 2.28-1.552 3.285-1.23 3.285-1.23.645 1.653.24 2.873.12 3.176.765.84 1.23 1.91 1.23 3.22 0 4.61-2.805 5.625-5.475 5.92.42.36.81 1.096.81 2.22 0 1.606-.015 2.896-.015 3.286 0 .315.21.69.825.57C20.565 22.092 24 17.592 24 12.297c0-6.627-5.373-12-12-12"/>
+<!-- Source: https://fontawesome.com/icons/github -->
```
dhimmel: Should we say "Modified from" rather than "Source" to indicate we've made modifications?

vincerubinetti:
> Is there an options to define these SVGs in one of our embedded HTML files? I wonder if that makes more sense then having them as separate files that clutter `images`? @vincerubinetti what's your thought on that?

SVGs can be embedded in HTML, but they have to be embedded in-place. I don't know of a way to define it once, say in the `<head>`, and then reference it multiple times elsewhere. I'll look into it and let you know.

content/images/github.svg
vincerubinetti: > `{.author_icon}`

Will wait for the `default-theme` PR to change this this, since without the `{height="13px" width="13px"}`, and since the `.author_icon` CSS class isn't defined yet, the svg will be huge: it'll scale to the full width of the page.

PR #167, opened 2019-01-25
Themes plugins refactor

vincerubinetti:
Rearranged setup of `/build` folder:

- `/themes` - modifications to pandoc output that are purely aesthetic, and don't affect functionality - usually CSS only
- `/plugins` - additions to pandoc output that give extra functionality, such as tooltips, menus, certain display fixes, etc. - usually javascript only
- `/assets` - other uncategorizable "config" options, such as the citation style

`pandoc-reference.docx` was moved to `/themes` since, to my understanding, that's really what it is: a reference for how pandoc should style the `.docx` output aesthetically, just like CSS is to the `.html` output.


slochower:
> pandoc-reference.docx was moved to /themes since, to my understanding, that's really what it is: a reference for how pandoc should style the .docx output aesthetically, just like CSS is to the .html output.

Correct.

PR #171, opened 2019-01-28
Create default css theme for styling elements (body, h1, figure, etc)

vincerubinetti:


build/themes/default.html
```diff
@@ -1,1093 +1,390 @@
 <style>
-/* From https://gist.github.com/dashed/6714393 with Manubot modifications */
 
-/*! normalize.css v2.1.3 | MIT License | git.io/normalize */
+    /* import google fonts */
+    @import url("https://fonts.googleapis.com/css?family=Open+Sans:400,600,700");
+    @import url("https://fonts.googleapis.com/css?family=Source+Code+Pro");
```
dhimmel: Do we have to worry about the dependency on Google? For example, doesn't china great :fire: :moyai: block google domains? Would it make sense to host these fonts on https://github.com/greenelab/manubot-resources/ (license permitting).

vincerubinetti: Do they completely block all google.com? I wasn't aware of that. Will they somehow block the whole html document if they see one link to google there?

If they just block the font import, then it will fall back to helvetica, and if the user doesn't have that installed, it'll fall back to whatever sans-serif they have (usually arial).

https://github.com/greenelab/manubot-rootstock/blob/e4b9996a3f9c53207596b9198b1524c2542a6835/build/themes/default.html#L14

dhimmel: So would it be possible to add the font files to `manubot-resources` and only use those if the googleapis import failed? It looks like they both have licenses that should allow this.

vincerubinetti: Actually it looks like github doesn't support this:
https://stackoverflow.com/questions/39693837/how-do-i-use-github-to-host-a-webfont

=/

dhimmel: I think a workaround would be to set GitHub pages to deploy the `master` branch of `manubot-resources`. One downside is that we wouldn't be able to provide versioning by commit hash then. Ideally, we would keep all links to repos versioned such that they are less likely to break.

dhimmel: Or on second thought, we could make a `gh-pages` branch that stores versioned files and we commit to not removing them. We use the custom domain `resources.manubot.org` such that we could even switch web hosts.

vincerubinetti: Would we be able to get the resource from `gh-pages` properly? Github can host fonts just fine, but when trying to link to the "raw" file, it doesn't work. According to that stackoverflow question, the reason is that gh raw files only support the text mime type.

vincerubinetti: Is the font issue blocking this merge? I'm thinking `gh-pages` won't solve the issue if regular branches can't provide the right mime type. We can always add in our own hosting location of the fonts later. The only change that would be required would be adding an extra `@import url()` statement below the google ones.

dhimmel: I think the GitHub Pages would properly serve the font files as it [seems to be](https://help.github.com/articles/mime-types-on-github-pages/) smart with mime types and is intended for hosting.

Not a blocker for merging, as we can always change this later. I think we potentially should, especially since some users do not like the omnipresence of google and would prefer not to use them if possible. In addition, to any potential issues with googleapis being blocked in china.

dhimmel:
@dongbohu can you review this?

build/themes/default.html
```diff
@@ -1,1093 +1,389 @@
...
```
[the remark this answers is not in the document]
vincerubinetti: We're trying to minimize the amount of file inclusion/movement/handling the user has to do. The idea was to have everything needed for one theme in one and only one file. In normal web development circumstances I would have much more separation of files and code, but in this case, where the user has to deal directly with the files locally and also (presumably) doesn't know anything about web development, we wanted the files to be as self-contained as possible. In fact everything except the images are embedded right into the (single) `.html` file.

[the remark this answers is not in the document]
dongbohu: :+1:

PR #178, opened 2019-02-08
Various aesthetic updates

vincerubinetti:
See file diff for changes. Too many and small changes to list here.

Hold off on merging this until plugins PR is finalized, in case I make more small adjustments to this.

### DIFF
--- a/build/themes/default.html
+++ b/build/themes/default.html
@@ -16,18 +16,23 @@
     }
 
     /* "page" element */
-    body {
-        position: relative;
-        box-sizing: border-box;
-        /* make resemble paper */
-        max-width: 8.5in;
-        font-size: 16px;
-        line-height: 1.5;
-        margin: 20px auto;
-        padding: 40px;
-        border-radius: 4px;
-        background: #ffffff;
-        box-shadow: 0 0 15px #bbbbbb;
+    @media only screen {
+        body {
+            position: relative;
+            box-sizing: border-box;
+            /* make resemble paper */
+            max-width: 8.5in;
+            font-size: 16px;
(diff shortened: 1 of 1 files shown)


vincerubinetti:
[will edit this comment with explanation of each change before review and merge]

build/themes/default.html
```diff
@@ -354,57 +367,69 @@
     /* -------------------------------------------------- */
     /* when printing */
     /* -------------------------------------------------- */
-
-    /* paper/page size and margins */
+    
     @page {
-        size: letter;
```
dhimmel: @vincerubinetti can you look at https://github.com/greenelab/manubot-rootstock/issues/105 / https://github.com/greenelab/manubot-rootstock/pull/107 to make sure removing letter won't cause a regression.

vincerubinetti: I can't seem to find a screenshot of the problem that that issue is referencing, but I'm almost certain that the issue was caused by the poor/dirty nature of the old CSS. The `<html>`, `<body>`, and `@page` had inconsistent and nonsensical margins and paddings. Recall that `@page` had something like `margin-left: 78mm; margin-top: 56mm; margin-right: 0; margin-bottom: 0`. This was most likely the issue with bottom text getting cutoff.

The new theme should work fine for any page size. I tested it with the various ISO 216 ("A") sizes, and it seems to work. Let's revisit this at merge time when we test all the plugins and new themes thoroughly amongst several different users and manuscripts.

dhimmel: Makes sense. I don't see why margins should only work with a single page size should things be properly implemented.

PR #179, opened 2019-02-08
Plugin suite

vincerubinetti:
This PR will contain the suite of front-end plugins for the HTML output.


dhimmel:
Is this ready for a first round of review from @dongbohu? I am thinking we want to get the existing plugins out as soon as possible so users can start trying out and benefiting from these great features!

vincerubinetti:
@dongbohu @dhimmel Dongbo and anyone in the lab can start testing the existing plugins that are here. When you do, I recommend you replace your `content/02.delete-me.md` file with [this content](https://pastebin.com/5i0BuDiC), which will give you a nice testing ground for all the different features.

build/plugins/anchors.html
```diff
@@ -1,15 +1,85 @@
+<!-- anchors plugin -->
```
dhimmel: Since we're replacing anchorsjs we can remove license note from readme.

build/build.sh
```diff
@@ -33,11 +33,19 @@ pandoc --verbose \
   --bibliography=$BIBLIOGRAPHY_PATH \
   --csl=$CSL_PATH \
   --metadata link-citations=true \
-  --mathjax \
   --include-after-body=build/themes/default.html \
```
dongbohu: This is probably more a question for @dhimmel:
I am little bothered by the fact that all these html files are actually JavaScript embedded in `<script>..</script>`. Wouldn't it be better to use `--include-in-header` option to include the JavaScript files directly?

dongbohu:
By the way, why `travis` failed?

build/build.sh
```diff
@@ -33,16 +33,19 @@ pandoc --verbose \
...
   --include-after-body=build/plugins/table-scroll.html \
   --include-after-body=build/plugins/anchors.html \
   --include-after-body=build/plugins/accordion.html \
   --include-after-body=build/plugins/tooltips.html \
   --include-after-body=build/plugins/jump-to-first.html \
   --include-after-body=build/plugins/ref-highlight.html \
+  --include-after-body=build/plugins/table-of-contents.html \
```
agitter: This is what is causing the Travis failure
```
File build/plugins/table-of-contents.html not found in resource path
```

build/plugins/accordion.html
```diff
@@ -0,0 +1,208 @@
...
+
+    (function() {
+        ///////////////////////////
+        // ASSETS
+        ///////////////////////////
+
+        // html svg of arrow icon
+        // modified from: https://fontawesome.com/icons/angle-down
+        let arrowIcon =
+            "<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 448 512'><path fill='currentColor' d='M207.029 381.476L12.686 187.132c-9.373-9.373-9.373-24.569 0-33.941l22.667-22.667c9.357-9.357 24.522-9.375 33.901-.04L224 284.505l154.745-154.021c9.379-9.335 24.544-9.317 33.901.04l22.667 22.667c9.373 9.373 9.373 24.569 0 33.941L240.971 381.476c-9.373 9.372-24.569 9.372-33.942 0z'></path></svg>";
```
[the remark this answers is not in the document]
vincerubinetti: I'm not sure how to elegantly do this, because most of the length comes from the long svg path `d=` parameter which defines the point of the path. I suppose I can split it up arbitrarily in the spaces of the path description, or if you know any automatic formatter tools to do this.

build/plugins/accordion.html
```diff
@@ -0,0 +1,208 @@
+<!-- accordion plugin -->
+
+<script>
```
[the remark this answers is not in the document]
vincerubinetti: I'm happy to incorporate all of the style changes except for this one. I find 2 to be impossible to read. Also in this case, all of the other Manubot code seems to be using 4. Further, this code is intended to be possibly be read by the users of Manubot, and I think 4 is more readable and widely accepted. We don't (and shouldn't) have so much logic nesting that 2 really saves much more space.

build/plugins/accordion.html
```diff
@@ -0,0 +1,208 @@
...
+                top: rect.top + margin.top + border.top,
+                right: rect.right + margin.right + border.right,
+                bottom: rect.bottom + margin.bottom + border.bottom
+            };
+        }
+
+        // ** for use with accordion plugin
+        // get closest element before specified element that matches query
+        function firstBefore(element, query) {
+            while (
```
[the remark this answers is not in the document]
vincerubinetti: The parenthesis were added by the formatter I was using (called `Prettier`). I'm in the process of switching over fully to eslint which should get rid of these issues. Note that I'm using the Greenelab `.eslintrc` config file, with the only modification being 2 -> 4 space indentation.

build/plugins/accordion.html
```diff
@@ -0,0 +1,208 @@
...
+                arrow.innerHTML = arrowIcon;
+                arrow.classList.add("icon_button", "accordion_arrow");
+                heading.insertBefore(arrow, heading.firstChild);
+
+                // attach click listener to heading and button
+                heading.addEventListener("click", onHeadingClick);
+                arrow.addEventListener("click", onArrowClick);
+
+                // collapse headings by default if on small screen, otherwise expand
+                if (isSmallScreen()) collapseHeading(heading);
```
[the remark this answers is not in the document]
vincerubinetti: This is also the unfortunate result of using `Prettier`. Switching to `eslint` per my other comment should fix this.

build/build.sh
vincerubinetti: I'm not sure what you mean by this.

The `--include-in-header` option of pandoc basically just copies and pastes text data into the `.html` output, verbatim. It doesn't hotlink things eg `<script src="http://googleapis.com/jquery.min.js"></script>`. I think the maintainer of pandoc actually did add a feature (at Daniel's request) to allow a URL reference in the `--include` parameters, but that still just fetches the resource at the URL and pastes it verbatim into the document, same as normal.

Also, some of our plugin files necessitate multiple `<script>` tags for configuration, for example Mathjax:

```
<script type="text/x-mathjax-config">
    // plugin configuration
</script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js?config=TeX-MML-AM_CHTML">
    // link to or code from MathJax or similar plugin
</script>
```
Even though our plugins will only ever need one `<script>` tag, I wanted to keep all the plugin files consistent in format and extension.

Daniel and I had talked about possibly moving our plugins to a hosted repo and hotlinking them in the `--include` parameters instead of having them directly in the manuscript repo, but they would still take the same form: `.html` files with `<script>` tags.

build/plugins/accordion.html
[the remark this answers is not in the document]
dongbohu: This is just a long string in JS. Please feel free to cut it anywhere you feel appropriate. I don't have any preferences once each line is shorted than 80 characters.

build/plugins/accordion.html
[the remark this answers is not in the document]
dongbohu: I never felt that 2 characters of indentation is impossible to read. If you strongly prefer indentation of 2 in this project, it is okay with me, expecially considering that it already includes some HTML/CSS/JS code with 4 chars of indentation. But for new projects such as `hetmech-frontend`, I do hope that we can stick to the rules specified in the our lab's Eslint config.

build/build.sh
dongbohu: I am not an expert of `pandoc`. If `--include-in-header` option doesn't help, then we don't any other options. :-P

vincerubinetti: It's weird to put javascript in an `.html` file, for sure. I don't know of any precedence for doing so. But I think we're in weirdly specific territory here, and I think under the circumstances it's the cleanest solution.

build/plugins/accordion.html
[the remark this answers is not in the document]
vincerubinetti: I have trouble reading 2 spaces, and sometimes even 4. I'm actually used to allman style brackets too, and still have trouble reading K&R. (I wish javascript was just bracket-less and white-space dependent like python).

I was planning on linting hetmech to 2 spaces, yes. I'll find someway to work on it in 4 and have it automatically convert to 2 on push.

build/plugins/accordion.html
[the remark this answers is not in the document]
dhimmel: > I don't have any preferences once each line is shorted than 80 characters.

Hmm. I'm not sure I'm in favor of such short line lengths. The 80 character max is really an outdated concept.

For these SVGs that we're tracking with git, I think it makes sense to have the entire points of the path on a single line. This seems like the most semantic way to break the line. Similar to how we do one-sentence-per-line in our markdown prose. I think we should avoid lines over 120 characters when possible, unless that requires breaking something line this `d=` list or a URL, etcetera.

build/plugins/accordion.html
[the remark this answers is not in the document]
dhimmel: I am more in favor of 2 spaces than 4 for this project, but don't have a strong preference.

build/plugins/accordion.html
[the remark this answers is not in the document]
vincerubinetti: 80 is the standard for javascript these days. However it generally doesn't apply to long continuous strings or comments, just logic. What we have here is a long string. Even though it contains structured content, I don't think it looks good to split up with `+`'s.

What about moving the assets outside of the script tag so they can just be plain html, like this:

```html

<script>
// DESCRIPTION
// this manubot plugin blah blah

// OPTIONS

...plugin code...
</script>

// modified from: https://fontawesome.com/icons/
<svg id="left_arrow_icon" viewBox="0 0 512 512">
    <path
        fill="blah"
        d="1 09287129 8712 172e91872e 97812 987"
    >
    </path>
</svg>
```

build/build.sh
```diff
@@ -33,53 +33,19 @@ pandoc --verbose \
   --bibliography=$BIBLIOGRAPHY_PATH \
   --csl=$CSL_PATH \
   --metadata link-citations=true \
-  --mathjax \
   --include-after-body=build/themes/default.html \
-  --include-after-body=build/plugins/analytics.html \
```
dhimmel: Was the intention to remove the analytics and require the user to put it back in the build command? There's not much harm keeping it enabled by default.

build/build.sh
```diff
@@ -33,53 +33,19 @@ pandoc --verbose \
...
-    --filter=pandoc-tablenos \
-    --bibliography=$BIBLIOGRAPHY_PATH \
-    --csl=$CSL_PATH \
-    --metadata link-citations=true \
-    --reference-doc=build/themes/default.docx \
-    --resource-path=.:content \
-    --output=output/manuscript.docx \
-    $INPUT_PATH
-fi
-
```
dhimmel: Must not delete PDF and DOCX build commands.

vincerubinetti: These were both mistakes. Had removed them to speed up the build during testing, forgot to add them back in. Fixed.

build/plugins/annotations.html
```diff
@@ -0,0 +1,29 @@
+<!-- annotations plugin configuration -->
```
dhimmel: I lean towards continuing to call this plugin "hypothesis" for backwards compatibility.

dhimmel:
@dongbohu, did you want to re-review after vince's new changes. I believe the lightbox plugin is new. @vincerubinetti, it may help to commit the latest changes to make sure dongbo reviews the most up-to-date version.

Looking to merge these plugins ASAP.

dongbohu:
I will review it in the afternoon.

build/plugins/annotations.html
vincerubinetti: I renamed the plugin file name back to `hypothesis.html`, but left the comment there to describe what it actually does for those who don't know. Is that okay?

I believe peoples' `hypothesis.html` file contents are going to have to be updated anyway because of added the configuration lines in it.

PR #187, opened 2019-03-12
Plugin suite bugfix & update

vincerubinetti:
*Lightbox*
Changed the image copy procedure (copying `` in document to lightbox ``) from shallow (new `` with same `src`) to deep copy (copy DOM node fully with all properties/attributes/styles, except for explicitly deleted ones like `width`, `height`, `id`, etc.).

*Tooltips*
Changed dimension styles a bit. Allowed tall tooltips to scroll vertically. Decreased spacing on nav bar.

*Table of Contents*
Allowed clicking off of the panel to close it, as expected on mobile.

### DIFF
--- a/build/plugins/accordion.html
+++ b/build/plugins/accordion.html
@@ -165,8 +165,8 @@
             let rect = {};
             rect.left = 0;
             rect.top = 0;
-            rect.right = window.innerWidth;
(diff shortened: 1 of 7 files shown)


vincerubinetti:
Replaced by #191

PR #191, opened 2019-03-14
update lightbox tooltips and toc plugins

vincerubinetti:
*Lightbox*
Changed the image copy procedure (copying `` in document to lightbox ``) from shallow (new `` with same `src`) to deep copy (copy DOM node fully with all properties/attributes/styles, except for explicitly deleted ones like `width`, `height`, `id`, etc.).

*Tooltips*
Changed dimension styles a bit. Allowed tall tooltips to scroll vertically. Decreased spacing on nav bar.

*Table of Contents*
Allowed clicking off of the panel to close it, as expected on mobile.

### DIFF
--- a/build/themes/default.html
+++ b/build/themes/default.html
@@ -105,6 +105,7 @@
     /* links */
     a {
         color: #2196f3;
+        word-break: break-all;
     }
 
     /* superscripts and subscripts */
@@ -484,7 +485,7 @@
     @media print {
         @page {
             /* suggested printing margin */
-            margin: 0.75in;
+            margin: 1in;
         }
 
         /* document and "page" elements */
@@ -710,12 +711,13 @@
             position: absolute;
             width: 50%;
             min-width: 240px;
+            max-width: 75%;
             z-index: 1;
         }
 
         /* tooltip content */
         #tooltip_content {
-            margin: 10px 0;
+            margin-bottom: 5px;
             padding: 20px;
             border-radius: 5px;
             border: solid 1px #bdbdbd;
@@ -729,13 +731,13 @@
         #tooltip_content > p,
         #tooltip_content > figure {
             margin: 0;
+            max-height: 320px;
+            overflow-y: auto;
         }
 
-        /* tooltip copy of figure captions */
-        #tooltip_content > figure > figcaption {
-            margin: 0;
-            margin-top: 20px;
-            padding: 0;
+        /* tooltip copy of <img> */
+        #tooltip_content > figure > img {
+            max-height: 260px;
         }
 
         /* navigation bar */
@@ -982,7 +984,7 @@
 
     @media only screen {
         /* regular <img> in document when hovered */
-        .lightbox_img:hover {
+        .lightbox_document_img:hover {
             cursor: pointer;
         }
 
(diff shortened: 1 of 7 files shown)


build/themes/default.html
```diff
@@ -484,7 +485,7 @@
     @media print {
         @page {
             /* suggested printing margin */
-            margin: 0.75in;
+            margin: 1in;
```
dhimmel: Oops looks like we changed the margins back to 1 inch from 0.75

vincerubinetti: Sorry, I forgot to make that change in my local master copy of the front-end since we did it quick on your computer. We can change that back today.

slochower:
This is not so much a bug as it is just a comment and question for @vincerubinetti  -- that's why I'm posting on this lightbox PR instead of in an Issue. I'm running an unsupported mix of Manubot features, including replacing the usual figure filter `pandoc-fignos` with `pandoc-crossref` due to the latter's subigure support.

See this manuscript for reference: https://slochower.github.io/smirnoff-host-guest-manuscript/v/a2fec1ee6bf387a1e664186f8ddd2c15facadcf7/

A tooltip that points to a plain figure, like Figure 1, works as expected. A tooltip that points to a combined subfigure, like Figure 2, does not work as expected -- the preview basically shows an embedded copy of the entire HTML page.

The subfigures are created like so:
```
<div id="fig:atom-names">
![](images/atom-names-trimer.png){width=3.5in}
![](images/gaff-atom-types.png){width=3.5in}

Atom names (left) and GAFF atom types (right) for a glucose monomer in αCD shown with two flanking monomers. The remaining three glucose monomers are hidden for clarity.
</div>
```
The subfigure tooltips *do* link to anchoring elements (`#...`) on the page that *do* resolve to the right place (although I'm not sure if this is due to a quirk of my setup or not, sometimes I have to refresh twice to go to the right place on the page). Note that I'm using one combined caption for both subfigures, not individual subfigure captions that would go between the `[]`s.

Just curious if you think adjusting the tooltips to deal with this would be easy and quick or not.

A slightly related point is that the figure and subfigure captions in the generated PDF are styled differently. The plain figure captions are reduced in size, but the subfigure captions are not, probably because they use different style markup (which you can check in the HTML, too).

slochower:
As a followup, the "jump to the position in the text" chevron/double up arrow (not sure what its real name is) is occasionally misplaced. I'm not sure whether that is related to using `pandoc-crossref` or not. Apologies for not having time to test at the moment -- since I'm the one running a custom mix of things, I ought to be troubleshooting.

image

(It is between 1 and $\mu$s which is written `200 snapshots over 1 $\mu$s` in the source).

And in other cases, the little chevron is missing for the subfigure captions altogether. I'm guessing this is because they are not styled as figure captions.

image

vincerubinetti:
@slochower I think both of these merit creating issues. When you have time, please create one issue for each of the two problems.

I know why they're happening, and they're happening because of assumptions made about the layout of the html. So far, we haven't been supporting/assuming people would add their own images "manually" (without fignos, I guess) as you've done here.

The unfortunate reality is that _some_ assumptions will always have to be made. But it might be possible in this case to support this edge case. I'd like to try to make the plugins as flexible as possible.

PR #193, opened 2019-03-15
fix auto option in toc plugin, fix printing page margin

vincerubinetti:
Fixes https://github.com/zach-hensel/low-noise-manuscript/commit/3ae5e37f79b31e06772a80a15ac00474d6d47330 and sets page margin back to `0.75in`.


zach-hensel:
FYI I think this turns isSmallScreen() into isntSmallScreen() and makes TOC close automatically on link clicking... instead making this change to line 58 works for me: 
`(options.startOpen === 'auto' && !isSmallScreen())`

vincerubinetti:
@zach-hensel Thanks for the tip. Will fix this with a PR today.

PR #194, opened 2019-03-18
change auto open/close option functionality and comments

vincerubinetti:
@zach-hensel Maybe you want to take a look at this since you caught the bug. I've changed the option name (because it applies to more than just whether the toc *starts* open or closed) and also the behavior a little bit.

I don't think we want to change the default to `'auto'` though, just to avoid cluttering the screen. Thoughts on this @dhimmel

### DIFF
--- a/build/plugins/table-of-contents.html
+++ b/build/plugins/table-of-contents.html
@@ -27,10 +27,10 @@
             charLimit: '50',
             // whether or not to show bullets next to each toc item
             bullets: 'false',
-            // whether to always start closed ('false'), always start opened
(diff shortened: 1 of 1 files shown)


dhimmel:
@zach-hensel, did you want to review the code changes? (I think you may have a better grasp on frontend development than me). If not, we can also have @dongbohu review.

dhimmel:
Thanks zh for helping us identify the issue in https://github.com/zach-hensel/low-noise-manuscript/commit/3ae5e37f79b31e06772a80a15ac00474d6d47330#r32772016

PR #195, opened 2019-03-18
Specify explicit size for icon svgs

vincerubinetti:
Sometimes you can see a 'flash' of a very large svg image while the page is loading. This is because the svg has loaded before the css specifying its size has. To fix, specify a reasonable size explicitly in the svg (so it will load at that size by default and then been overridden by whatever size the css tells it to be.


dhimmel:
Is removing `xmlns="http://www.w3.org/2000/svg"` related to the sizing or an unrelated change?

vincerubinetti:
> Is removing `xmlns="http://www.w3.org/2000/svg"` related to the sizing or an unrelated change? If unrelated, can you provide a justification we can use for the commit message.

That was a mistake. Re-added.

dhimmel:
https://github.com/manubot/resources/commit/0496a7e31dca2f2858a7e6440b8c557108c41c14#diff-e9c9ce7929c050b9071f8bb360929d30 removes the licensing attribution in the comment.

content/images/orcid.svg
```diff
@@ -1,4 +1,4 @@
 <!-- Modified from: https://figshare.com/articles/ORCID_iD_icon_graphics/5008697 -->
```
dhimmel: This also needs to be updated to comment from https://github.com/manubot/resources/pull/1

content/images/github.svg
```diff
@@ -1,4 +1,4 @@
-<!-- Modified from: https://fontawesome.com/icons/github -->
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 496 512">
+<!-- Modified from: https://fontawesome.com/icons/github under its CC BY 4.0 License -->
```
dhimmel: Nitpik: does it make more sense to remove the colon since it breaks up the sentence. Now might be a good time to make this change.

vincerubinetti: I think you're right since we now have the license after the url as part of the same sentence. Changed it.

Issue #192, opened 2019-03-15
Increasing the density of PDF exports to reduce number of pages


dhimmel:
Manubot has never been good at producing space-efficient PDFs. The recent frontend updates have brought a much better experience for the HTML in-browser view. However, I think they have exacerbated the space inefficiency of PDF exports a bit. For example, see the current:

+ [rephetio manuscript pdf](https://git.dhimmel.com/rephetio-manuscript/v/b7f9f14047594c835bfeaf3368cd35924e4eac31/manuscript.pdf) now 71 pages 
+ [meta-review pdf](https://greenelab.github.io/meta-review/v/6e7458dcb205a3d76daf25ee08236c24dc4de66e/manuscript.pdf) now 27 pages

My main motivation for decreasing the PDF page number is that many people print out PDFs, so it makes sense to be as efficient as possible. In general, I think most readers prefer space efficient PDFs, but I could be wrong here.

Anyways, there are things we can play with like font sizes and margins. While we probably won't be able to achieve the space efficiency of a heavily typeset document [like this](https://www.biorxiv.org/content/biorxiv/early/2019/03/14/578674.full.pdf), @vincerubinetti are there some low hanging fruit?

vincerubinetti:
First, I'd say that I think those are very nicely readable, and are what I would expect as an average reader. That is, those are inherently long papers that you've written. However, I'm not an academic, and that population might find it not in line with expectation.

I still think we have a good default setting, assuming you want Manubot to be targeted to any type of writers, not just academic papers. I was intending to make another theme that would be more compact (probably called `compact.html`). In it, we could apply the following styles, which would help:

```CSS
/* decrease page margin */
@page { margin: 0.5in; }
/* decrease font size and line spacing */
body { font-size: 10pt; line-height: 1.25; }
/* decrease table cell padding */
th, td { padding: 5px; }
/* decrease paragraph, figure, table, etc margins */
p, hr, figure, table, pre { margin-top: 20px; margin-bottom: 20px; }
/* decrease heading margins */
h1, h2, h3, h4, h5, h6 { margin: 10px 0; }
```

You can test out these modifications in your dev console to see how they look.

Additionally I had intended to make a two-column layout theme as well, though that might be more involved than the above.

dhimmel:
[...]

PR #201, opened 2019-04-04
compactify print view

vincerubinetti:
@dhimmel @slochower @agitter Per request (#192), I've made changes to the print CSS rules to reduce the number of pages when printing. This includes the following:

- smaller overall font size
- smaller line spacing
- smaller page margins (0.75in -> 0.5in)
- smaller heading margins
- smaller table cell padding
- smaller button icons

See the diff for more details.

Here's how it looks on the meta-review:
[meta-review.pdf](https://github.com/manubot/rootstock/files/3044822/meta-review.pdf)

### DIFF
--- a/build/themes/default.html
+++ b/build/themes/default.html
@@ -67,8 +67,9 @@
         font-weight: bold;
     }
 
-    /* header and biggest heading */
+    /* biggest heading */
     h1 {
+        margin: 40px 0;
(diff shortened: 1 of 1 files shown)


vincerubinetti:
I also reduced the size of the figure captions as it looked weird having the table captions small but not the figure captions.

Issue #170, opened 2019-01-28
Consider switching from weasyprint to another pdf converter


vincerubinetti:
I noticed that manubot used to use [wkhtmltopdf](https://github.com/greenelab/manubot-rootstock/issues/77) but switched to weasyprint due to the former lacking svg capabilities.

I've just run into a deficiency in weasyprint, in that it [doesn't support](https://github.com/Kozea/WeasyPrint/issues/357) the CSS [`calc()` value](https://caniuse.com/#feat=calc). We don't absolutely need to use the `calc()` value, at least right now. Conceivably in the future, it could be required to achieve a certain look/layout in a "clean" way compliant with modern best practices.

I'm wary that weasyprint reimplements its own new rendering engine, instead of using a well established standard like webkit or gecko. Even if weasy somehow does basic rendering as good as or better than webkit/gecko (projects that have hundreds of contributors), I think it will necessarily miss some of the more advanced functionality of CSS that we might want/need to use. This `calc()` value for example isn't exactly a cutting edge CSS feature; it came into being around 2011.

Again, it's a small problem, but it may be worth it to research tools like [athena](https://www.athenapdf.com/) and [prince](https://www.princexml.com/doc-prince/), and see if they're more fully featured (preferably based on webkit) than weasyprint, and would work with pandoc and our build process.

dhimmel:
Sure happy to consider open source alternatives to weasyprint.

I can't speak regarding the capabilities of athena or prince, but I've been pretty satisfied with the functionality of weasyprint and exceptionally happy with the support. Whenever we report issues, we've responses quickly... even when their not really weasyprint issues.

Not to say we shouldn't evaluate alternatives if there are important missing features. It probably makes sense to revisit this issue more actively if we decide to switch to a manubot Docker image that could more readily include dependencies that are not already part of conda-forge.

dhimmel:
Random links to look into more:

- https://github.com/RelaxedJS/ReLaXed

dhimmel:
I tried `athenapdf` via it's [Docker](https://hub.docker.com/r/arachnysdocker/athenapdf) to create a PDF of the [bitcoin-whitepaper](https://git.dhimmel.com/bitcoin-whitepaper/v/ebfce2b530b141764275093eee8f424b18144420/):

```sh
docker run --rm \
  --volume=`pwd`:/converted/ \
  arachnysdocker/athenapdf:2.16.0 \
  athenapdf \
  --delay=2000 \
[...]

PR #210, opened 2019-04-17
Use Athena to build PDF, if Docker is installed

dhimmel:
Closes https://github.com/manubot/rootstock/issues/170

More information about Athena is available at
https://github.com/arachnys/athenapdf

Apply feature from meta-review commit:
https://github.com/greenelab/meta-review/commit/cffcfeaea0b6859051d16ab1e263d8bcf62a24b1

This commit was created with the command:
```
curl https://github.com/greenelab/meta-review/commit/cffcfeaea0b6859051d16ab1e263d8bcf62a24b1.patch | git apply
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 dist: trusty
 sudo: false
 language: generic
+services:
+  - docker
 branches:
   only:
     - master
--- a/build/build.sh
+++ b/build/build.sh
@@ -50,9 +50,12 @@ pandoc --verbose \
   --output=output/manuscript.html \
   $INPUT_PATH
 
+# Return null if docker command is missing, otherwise return path to docker
+DOCKER_EXISTS=`command -v docker`
+
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
-if [ "$BUILD_PDF" != "false" ]; then
-  echo "Exporting PDF manuscript"
+if [ "$BUILD_PDF" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
+  echo "Exporting PDF manuscript using WeasyPrint"
   if [ -L images ]; then rm images; fi  # if images is a symlink, remove it
   ln -s content/images
   pandoc \
@@ -71,6 +74,22 @@ if [ "$BUILD_PDF" != "false" ]; then
     --output=output/manuscript.pdf \
     $INPUT_PATH
   rm images
+fi
+
+# Create PDF output (unless BUILD_PDF environment variable equals "false")
+if [ "$BUILD_PDF" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
+  echo "Exporting PDF manuscript using Docker + Athena"
+  if [ -d output/images ]; then rm -rf output/images; fi  # if images is a directory, remove it
+  cp -r -L content/images output/
+  docker run \
+    --rm \
+    --volume `pwd`/output:/converted/ \
+    --security-opt seccomp:unconfined \
+    arachnysdocker/athenapdf:2.16.0 \
+    athenapdf \
+    --delay=2000 \
+    manuscript.html manuscript.pdf
+  rm -rf output/images
 fi
 
 # Create DOCX output (if BUILD_DOCX environment variable equals "true")
--- a/build/build.sh
+++ b/build/build.sh
@@ -80,7 +80,7 @@ fi
 if [ "$BUILD_PDF" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
   echo "Exporting PDF manuscript using Docker + Athena"
   if [ -d output/images ]; then rm -rf output/images; fi  # if images is a directory, remove it
-  cp -r -L content/images output/
+  cp -R -L content/images output/
   docker run \
     --rm \
     --volume `pwd`/output:/converted/ \


dhimmel:
Would be good to get a macOS test. @slochower do you use mac and does the Docker control flow work, like:
```
DOCKER_EXISTS=`command -v docker`
````

**Comment by slochower** [comment]
Yep.

```
Wed Apr 17 15:08:22 Triceratops:~
$ command -v docker
/usr/local/bin/docker
```

**Comment by agitter** [comment]
We could use Travis CI to test macOS without Docker.  They support macOS but [not Docker for macOS](https://docs.travis-ci.com/user/docker/).

This new build script may also bring us closer to Windows compatibility, working around some issues in #144.

**Comment by dhimmel** [comment]
> We could use Travis CI to test macOS without Docker

With the current design of roostock, I don't think this makes sense, since it will double the number of CI builds for manuscript instances. However, we perhaps should look into macOS builds for https://github.com/manubot/manubot. Or think about future ways we can create a testing manuscript that performs additional tests beyond the template of rootstock.

> This new build script may also bring us closer to Windows compatibility

@agitter if WeasyPrint and its `pango` dependency is the sole thing preventing Windows environment compatibility, does your calculus from https://github.com/manubot/rootstock/issues/170#issuecomment-484182440 change. I.e. would you want to require docker for PDF creation?

**Comment by agitter** [comment]
> it will double the number of CI builds for manuscript instances

Sorry, I didn't explain myself well enough.  I was thinking that we could create a temporary branch / pull request for one time testing of this build script in a clean macOS environment on Travis.  If anyone else is willing to test locally on macOS without Docker, that wouldn't be necessary.

Despite my comment above, Windows support doesn't change my thoughts in #170 much regarding requiring Docker for PDF builds.  Because we haven't confirmed Manubot setup works in Windows with Docker, I expect local PDF builds on Windows are going to be in limited demand.  I am assuming there are more Linux / macOS users who would want local PDFs without Docker than there are Windows users.

My understanding is that couldn't have local PDF builds on Windows unless we remove the WeasyPrint dependency from the environment and require Docker.

**Comment by dhimmel** [comment]
>  I am assuming there are more Linux / macOS users who would want local PDFs without Docker than there are Windows users.

Okay we can keep WeasyPrint, with the cost of not being able to create a windows-compatible `environment.yml`.

> Don't have a Mac without Docker, to test what happens then

Can you just change 

```
DOCKER_EXISTS=`command -v docker`
```

to

```
DOCKER_EXISTS=`command -v dockerxxx`
```

And then see whether the the WeasyPrint PDF creation executes and succeeds, and that the Athena PDF creation does not execute?

**Comment by slochower** [comment]
Looks good. Do get one warning, though:

```
cp: the -H, -L, and -P options may not be specified with the -r option.
```

Maybe we want `cp -R -L`. The man page does specify `-R` for recursive, but I always use `-r` from habit on linux and it must work through some compatibility.

**Comment by dhimmel** [comment]
Looks like the [linux command](http://man7.org/linux/man-pages/man1/cp.1.html), `-R`, `-r`, or `--recursive`. However, macOS only [supports](https://ss64.com/osx/cp.html) `-R`. So yeah we should switch this.

**Comment by dhimmel** [comment]
@agitter any remaining requests on your end? It wasn't clear whether you wanted addition macOS testing or whether slochower's empirical testing is enough.

**Comment by slochower** [comment]
I'm getting some issues with Athena. I'm wondering if it might be related to the plugins. I've asked to get more output from the build process so I can debug what's going on. Ref: https://github.com/arachnys/athenapdf/issues/195

**Comment by slochower** [comment]
Tried to eliminate plugins as a cause. Rendering the HTML like so still causes Athena to crash:

```
pandoc --verbose \
  --from=markdown \
  --to=html5 \
  --filter=pandoc-eqnos \
  --filter=pandoc-tablenos \
  --filter=pandoc-fignos \
  --bibliography=$BIBLIOGRAPHY_PATH \
  --csl=$CSL_PATH \
    --output=output/manuscript.html \
  $INPUT_PATH

  # --include-after-body=build/themes/default.html \
  # --include-after-body=build/plugins/table-scroll.html \
  # --include-after-body=build/plugins/anchors.html \
  # --include-after-body=build/plugins/accordion.html \
  # --include-after-body=build/plugins/tooltips.html \
  # --include-after-body=build/plugins/jump-to-first.html \
  # --include-after-body=build/plugins/link-highlight.html \
  # --include-after-body=build/plugins/table-of-contents.html \
  # --include-after-body=build/plugins/lightbox.html \
  # --mathjax \
  # --variable math="" \
  # --include-after-body=build/plugins/math.html \
  # --include-after-body=build/plugins/hypothesis.html \
  # --include-after-body=build/plugins/analytics.html \

```

slochower:
@dhimmel to help the Athena folks with debugging, I would have thought to point them to `manuscript.html` on the `output` GitHub branch, but that of course doesn't have the images or plugins locally. Is it `index.html` on the `gh-pages` branch?

dhimmel:
I see in the [Travis logs](https://travis-ci.org/slochower/smirnoff-host-guest-manuscript/builds/530392091#L2296) that the athena command failed with the message `The renderer process has crashed.` 

> @dhimmel to help the Athena folks with debugging, I would have thought to point them to manuscript.html on the output GitHub branch

You are right that athena is actually being run on `output/manuscript.html`. However, there should be very few differences to index.html on the `gh-pages` branch / `webpage` directory. In your case, I think [this versioned GitHub Pages URL](https://slochower.github.io/smirnoff-host-guest-manuscript/v/5c9b6058fb44adba5c97a646ebccc801d566393e/) is sufficient for athena debugging. When I enter it at https://www.athenapdf.com/ I get the error message:

> There was a problem converting this page, please try again or try another URL

slochower:
@dhimmel note: https://github.com/arachnys/athenapdf/issues/195#issuecomment-491075361

Unless you have qualms, I'll make a PR to `--shm-size="2g"` to `build.sh`.

PR #220, opened 2019-05-09
Increase shared memory (--shm-size) to prevent athenapdf Docker from crashing

slochower:
This PR is to address situations where the Electron renderer that Athena uses to produce PDFs crashes. There are at least two ways to avoid this: either by passing `--shm-size="2g"` to `docker run` or by mounting `/dev/shm`, both of which potentially work on Travis. The former has been tested on a personal repository and seems to alleviate the crashes.

Refs https://github.com/arachnys/athenapdf/issues/195

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -82,10 +82,10 @@ if [ "$BUILD_PDF" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
   if [ -d output/images ]; then rm -rf output/images; fi  # if images is a directory, remove it
   cp -R -L content/images output/
   docker run \
-    --shm-size="2g" \
     --rm \
-    --volume `pwd`/output:/converted/ \
-    --security-opt seccomp:unconfined \
+    --shm-size=1g \
+    --volume=`pwd`/output:/converted/ \
+    --security-opt=seccomp:unconfined \
     arachnysdocker/athenapdf:2.16.0 \
     athenapdf \
     --delay=2000 \
--- a/build/build.sh
+++ b/build/build.sh
@@ -82,6 +82,7 @@ if [ "$BUILD_PDF" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
   if [ -d output/images ]; then rm -rf output/images; fi  # if images is a directory, remove it
   cp -R -L content/images output/
   docker run \
+    --shm-size="2g" \
     --rm \
     --volume `pwd`/output:/converted/ \
     --security-opt seccomp:unconfined \


dhimmel:
[Docker run docs](https://docs.docker.com/engine/reference/run/) for `--shm-size`:

> Size of `/dev/shm`. The format is `<number><unit>`. number must be greater than 0. Unit is optional and can be b (bytes), k (kilobytes), m (megabytes), or g (gigabytes). If you omit the unit, the system uses bytes. If you omit the size entirely, the system uses 64m.

From "[What Is /dev/shm And Its Practical Usage](https://www.cyberciti.biz/tips/what-is-devshm-and-its-practical-usage.html)":

> shm / shmfs is also known as tmpfs, which is a common name for a temporary file storage facility on many Unix-like operating systems. It is intended to appear as a mounted file system, but one which uses virtual memory instead of a persistent storage device.

dhimmel:
From https://github.com/arachnys/athenapdf/issues/195#issuecomment-491079569 by @slochower:

> In my case, I think the default macOS docker memory allocation is 2 GB

My understanding is that by default Docker containers can use unlimited memory or cpus but that this can be modified by setting the commands `--memory` or `--cpus` during `docker run`. However, if we think that our container is limited to 2 GB of memory on some systems, we should probably reduce `--shm-size` to `1.5g` to give the rest of the container some memory to operate with.

@slochower what do you think?

It looks like our [Travis CI jobs](https://docs.travis-ci.com/user/reference/overview/) are allocated 7.5 GB of memory, so I think we should be fine in that regard.

slochower:
> My understanding is that by default Docker containers can use unlimited memory or cpus but that this can be modified by setting the commands --memory or --cpus during docker run. 

I don't recall ever changing the settings on my Mac, and this is what I see (so I think it's the default). I haven't looked into the situation on linux yet. 

image

I agree we can increase the `shm` size on Travis.

> we should probably reduce --shm-size to 1.5g to give the rest of the container some memory to operate with.

Is that how it works? If you allocate `--shm-size` it takes away from other `docker` processes?

dhimmel:
> Is that how it works? If you allocate `--shm-size` it takes away from other `docker` processes?

Not sure. Let me discuss this with @dongbohu

dongbohu:
I would prefer 1.0 GB of `shm-size`, at most 1.5 GB.

dhimmel:
@slochower according to Dongbo, the reason Mac and Windows have a memory limit (for all running Docker containers) is that they use a VirtualBox virtual machine to support Docker. However, on Linux the containers can directly run on the operating system.

I will set to 1g and merge.

slochower:
I see, thanks.

dhimmel:
@slochower actually if you have a chance, can you see if the `docker run` command from https://github.com/manubot/rootstock/pull/220/commits/3183dca7c15fc8e3e2c36224b93afa3582aba5e4 runs on your mac? I've confirmed it works on Linux, but want to make sure, especially with the `=` additions.

slochower:
Works. macOS 10.14 with Docker version 1.13.1, build 092cba3.

I see:
```
Exporting PDF manuscript using Docker + Athena
ATTENTION: default value of option force_s3tc_enable overridden by environment.
Converted 'file:///converted/manuscript.html' to PDF: 'manuscript.pdf'
PDF Conversion: 8285.595ms
Build complete
```

I've gotten that `ATTENTION` warning before on my own stuff (I ran this on the bundled `rootstock` default manuscript). I think it is an Electron thing. Not clear how to debug.

slochower:
Based on [this recording](https://asciinema.org/a/WQZgZ8EAadGw5DMp4tTSM9iwq), building the bundled `rootstock` repository manuscript uses about 180 MB of Docker memory.

dhimmel:
@slochower we can safely ignore `ATTENTION ... force_s3tc_enable` as per https://github.com/nwjs/nw.js/issues/5335, but not sure if there's an easy fix to stop it from appearing.

What do you mean by "building the bundled rootstock repository manuscript"? Are you saying you created a Docker environment with the rootstock environment? Or are you just talking about running the `athenapdf` Docker image?

slochower:
> What do you mean by "building the bundled rootstock repository manuscript"? Are you saying you created a Docker environment with the rootstock environment? Or are you just talking about running the athenapdf Docker image?

Building the manuscript for _this_ repository, just to be clear I wasn't building the original manuscript that led to this PR, which is bigger & more complicated.

PR #222, opened 2019-05-17
Use 90. prefix for back-matter.md

dhimmel:
Closes https://github.com/manubot/manubot/issues/106

Enables creating two-digit prefixed content markdown files that will sort to come after back-matter. While back matter is intended to be the final markdown file, some users may want to place content after it.


slochower:
Yup, looks great.

Issue #229, opened 2019-06-06
`build.sh` stops abruptly without error when docker is missing


michaelmhoffman:
```
(manubot) mhoffman@mhoffman2 ~/src/collab/rootstock
$ build/build.sh 
Retrieving and processing reference metadata
## INFO
Manuscript content parts:
00.front-matter
01.abstract
02.delete-me
90.back-matter
## INFO
8 unique citations strings extracted from text
8 unique standard citations    
## INFO
requests-cache starting with 10 cached responses
## INFO
requests-cache finished with 10 cached responses
## INFO
Using UTC timezone.
Dating manuscript with the current datetime: 2019-06-06T20:52:02.890987+00:00
## INFO
Reading user-provided templating variables complete:
{}
## INFO
Generated manscript stats:
{
  "reference_counts": {
    "doi": 2,
    "isbn": 1,
    "pmcid": 1,
    "pmid": 1,
    "url": 2,
    "wikidata": 1,
    "total": 8
  },
  "word_count": 1194
}
Exporting HTML manuscript

(manubot) mhoffman@mhoffman2 ~/src/collab/rootstock [1]
$ echo $?
1
```

I expected to get a PDF.

I assume it is ending early because I don't have `docker` on my path.

dhimmel:
Interesting. On the systems we tested on, the following line worked when docker was missing:

https://github.com/manubot/rootstock/blob/b9e1d42b51a1b6a2807ce82840c49f1791f65ac5/build/build.sh#L53-L54

It would return nothing. But assuming that https://github.com/manubot/rootstock/pull/230 fixes the issue for you, it looks like

```
`command -v docker`
```

doesn't work on your system. What OS were you using and what terminal?

michaelmhoffman:
@agitter's explains in #230 how `set -o errexit` causes an abrupt termination here.

dhimmel:
@michaelmhoffman confirming the issue occurred on windows, where presumably the `command` command does not exist?

dongbohu:
To elaborate the issue, on a system that doesn't have `docker` installed, the return code of 
```
DOCKER_EXISTS=`command -v docker` 
``` 
is 1, which is an error, so the program will terminate due to `set -o errexit` at the beginning of `build.sh`.

dhimmel:
[...]

PR #230, opened 2019-06-06
Fix `DOCKER_EXISTS` setting so it doesn't fail when docker missing.

michaelmhoffman:
Also move to recommended quoted $() instead of backticks.

Fixes manubot/rootstock#229

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -51,7 +51,7 @@ pandoc --verbose \
   $INPUT_PATH
 
 # Return null if docker command is missing, otherwise return path to docker
-DOCKER_EXISTS=`command -v docker`
+DOCKER_EXISTS="$(command -v docker || true)"
 
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
 if [ "$BUILD_PDF" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then


dhimmel:
Looping @dongbohu in here since my shell scripting knowledge is limited.

build/build.sh
```diff
@@ -51,7 +51,7 @@ pandoc --verbose \
   $INPUT_PATH
 
 # Return null if docker command is missing, otherwise return path to docker
-DOCKER_EXISTS=`command -v docker`
+DOCKER_EXISTS="$(command -v docker || true)"
```
dhimmel: So this is saying that if the `command -v docker` returns null, then set `DOCKER_EXISTS=true`? Seems like this will always make `DOCKER_EXISTS` not null and therefore trigger the docker PDF build below?

agitter: @dhimmel there is some relevant discussion here: https://unix.stackexchange.com/questions/325705/why-is-pattern-command-true-useful/325727
> Rephrased: it masks the error status of the previous command.

In some quick testing with Git Bash on Windows, this is needed for a script with `set -o errexit`.
```
DOCKER_EXISTS=`command -v docker`
```
stops the script if `docker` isn't available, as reported in #229.

dhimmel: Okay here is what I get using bash on Linux:

```
$ DOCKER_EXISTS=`command -v docker || true`
$ echo $DOCKER_EXISTS
/usr/bin/docker
$ DOCKER_EXISTS=`command -v docker-mispelled || true`
$ echo $DOCKER_EXISTS

$ DOCKER_EXISTS=`command-mispelled -v docker || true`
command-mispelled: command not found
$ echo $DOCKER_EXISTS

```

So it does seem that `|| true` has the intended effect of silencing the error.

Is the switch from backticks to `$()` for command substitution required for windows bash, or is this a preference, as per [this discussion](https://stackoverflow.com/a/24592916/4651668)?

dhimmel: I wonder if we should use:

```sh
DOCKER_EXISTS=$(docker --version || true)
```

Because if the `docker` command exists, but the `command` command doesn't, we still want to make `DOCKER_EXISTS` non-null.

dongbohu: @dhimmel: `$(..)` was originally introduced in ksh, and later bash and POSIX borrowed this idea because it was better to read and nest other commands inside. I rarely used this format because it used to be incompatible with classic shell. But since it is now in POSIX, it is much safer now.

agitter: Using `docker --version` instead of `command` does seem more direct and relevant here.

dongbohu: When using `DOCKER_EXISTS=$(docker --version || true)` on a system that doesn't have docker installed, you may get error messages in stderr like this:
```
 Command 'docker' not found, but can be installed with:

sudo snap install docker     # version 18.06.1-ce, or
sudo apt  install docker.io

See 'snap info docker' for additional versions.
```
This is distracting. You have to hide it by this:
`DOCKER_EXISTS=$(docker --version > /dev/null 2>&1 || true)`
This may not be what you like.

`command` is actually a bash built-in:
https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html#Bash-Builtins

dhimmel:
Draft commit message for the merge:

```
build.sh: set DOCKER_EXISTS with a zero exit status

Merges https://github.com/manubot/rootstock/pull/230
Closes https://github.com/manubot/rootstock/issues/229

Previously, `command -v docker` returned a non-zero exit status when
docker did not exist. Because build.sh uses `set -o errexit`, this caused
the build script to exit. By appending `|| true`, the exit status will always
be zero.

Use recommended $() instead of backticks for command substitution.
https://stackoverflow.com/a/24592916/4651668
```

PR #233, opened 2019-06-10
Harden shell scripts with improved robustness & verbosity

dhimmel:
Writing good shell scripts that provide useful debugging information seems to be challenging. This PR attempts to adopt best practices as [discussed here](https://kvz.io/blog/2013/11/21/bash-best-practices/).

@dongbohu @michaelmhoffman @agitter feel free to suggest any other ways our shell scripts could be improved.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -56,7 +56,7 @@ pandoc --verbose \
 DOCKER_EXISTS="$(command -v docker || true)"
 
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
-if [ "${BUILD_PDF:-true}" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
+if [ "$(BUILD_PDF:-true)" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
   echo "Exporting PDF manuscript using WeasyPrint"
   if [ -L images ]; then rm images; fi  # if images is a symlink, remove it
   ln -s content/images
@@ -79,7 +79,7 @@ if [ "${BUILD_PDF:-true}" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
 fi
 
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
-if [ "${BUILD_PDF:-true}" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
+if [ "$(BUILD_PDF:-true)" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
   echo "Exporting PDF manuscript using Docker + Athena"
   if [ -d output/images ]; then rm -rf output/images; fi  # if images is a directory, remove it
   cp -R -L content/images output/
@@ -96,7 +96,7 @@ if [ "${BUILD_PDF:-true}" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
 fi
 
 # Create DOCX output (if BUILD_DOCX environment variable equals "true")
-if [ "${BUILD_PDF:-false}" = "true" ]; then
+if [ "$(BUILD_PDF:-false)" = "true" ]; then
   echo "Exporting Word Docx manuscript"
   pandoc --verbose \
     --from=markdown \
--- a/build/build.sh
+++ b/build/build.sh
@@ -1,4 +1,6 @@
-set -o errexit
+set -o errexit \
+    -o nounset \
+    -o xtrace
 
 # Set timezone used by Python for setting the manuscript's date
 export TZ=Etc/UTC
@@ -54,7 +56,7 @@ pandoc --verbose \
 DOCKER_EXISTS="$(command -v docker || true)"
 
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
-if [ "$BUILD_PDF" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
+if [ "${BUILD_PDF:-true}" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
   echo "Exporting PDF manuscript using WeasyPrint"
   if [ -L images ]; then rm images; fi  # if images is a symlink, remove it
   ln -s content/images
@@ -77,7 +79,7 @@ if [ "$BUILD_PDF" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
 fi
 
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
-if [ "$BUILD_PDF" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
+if [ "${BUILD_PDF:-true}" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
   echo "Exporting PDF manuscript using Docker + Athena"
   if [ -d output/images ]; then rm -rf output/images; fi  # if images is a directory, remove it
   cp -R -L content/images output/
@@ -94,7 +96,7 @@ if [ "$BUILD_PDF" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
 fi
 
 # Create DOCX output (if BUILD_DOCX environment variable equals "true")
-if [ "$BUILD_DOCX" = "true" ]; then
+if [ "${BUILD_PDF:-false}" = "true" ]; then
   echo "Exporting Word Docx manuscript"
   pandoc --verbose \
     --from=markdown \
--- a/build/build.sh
+++ b/build/build.sh
@@ -12,7 +12,7 @@ export TZ=Etc/UTC
 export LC_ALL=en_US.UTF-8
 
 # Generate reference information
-echo "Retrieving and processing reference metadata"
+>&2 echo "Retrieving and processing reference metadata"
 manubot process \
   --content-directory=content \
   --output-directory=output \
@@ -29,7 +29,7 @@ mkdir -p output
 
 # Create HTML output
 # http://pandoc.org/MANUAL.html
-echo "Exporting HTML manuscript"
+>&2 echo "Exporting HTML manuscript"
 pandoc --verbose \
   --from=markdown \
   --to=html5 \
@@ -61,7 +61,7 @@ DOCKER_EXISTS="$(command -v docker || true)"
 
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
 if [ "${BUILD_PDF:-true}" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
-  echo "Exporting PDF manuscript using WeasyPrint"
+  >&2 echo "Exporting PDF manuscript using WeasyPrint"
   if [ -L images ]; then rm images; fi  # if images is a symlink, remove it
   ln -s content/images
   pandoc \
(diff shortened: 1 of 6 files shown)


michaelmhoffman:
This is the template I use for shell scripts:

```
#!/usr/bin/env bash

## script.sh: <short description here>

## Copyright 2019 Michael M. Hoffman <michael.hoffman@utoronto.ca>

set -o nounset -o pipefail -o errexit

if [[ $# != 0 || "${1:-}" == "--help" || "${1:-}" == "-h" ]]; then
    echo usage: "$0"
    exit 2
fi
```

Rationale/discussion:

1. I believe there are several bashisms in this script already. I think it's fine to use bashisms because it is hard to imagine running this in an environment where bash isn't already there. But if so you should specify that it is bash at the top. (Also use `bash build/script.sh` instead of `sh build/script.sh` because the latter isn't going to work for some versions of `sh`.)

2. I recommend `set -o pipefail` because it will catch some things you want to be caught by `set -o errexit` but are masked.

3. Checking arguments and providing usage information is a good idea too.

4. I do not recommend `set -x`/`set -o xtrace`. Producing that level of noise by default violates best practices of Unix command-line interfaces, and leads to alert fatigue and people ignoring important information. For troubleshooting you can always run `bash -x script.sh`. As I use them, `set -o nounset -o pipefail -o errexit` are really failsafes for edge cases a programmer didn't consider rather than an expected path. You can also use `trap ... ERR` or `trap ... EXIT` to do something additional in case of unexpected failure (like suggesting running the script with `bash -x`).

Additional comments on shell style that jump out at me:

5. You used `${...}` in this patch in many places where you probably meant to use `$(...)`, I don't think it will work in its current form, and I'm surprised it passed CI.

6. It is good practice to almost always use double-quotes around parameter expansion to avoid any problems with weird characters. For example `--bibliography=$BIBLIOGRAPHY_PATH` --> `--bibliography="$BIBLIOGRAPHY_PATH"`

7. `...PATH` is usually used for a colon-separated list of directories, not a single file or directory

8. You may want some of these variables to be set only in cases where the user hasn't set and exported them. You can do that with a formulation like `TZ="${TZ:-Etc/UTC}`. I find overriding user settings of `TZ` and `LC_ALL` rather aggressive. If the latter is only to change Python settings you probably want to use `PYTHONIOENCODING` instead or even better fix it in the Python script itself.

9. I would advise adding [shellcheck](https://www.shellcheck.net/) to your editor and your CI workflow which would catch some of the errors above.

10. Debug info should go to stderr, not stdout. You want to use `echo >&2 "..."`. If the message isn't hard-coded and you are doing any shell expansion, `printf` is recommended instead.

11. You don't need `["${BUILD_PDF:-true}" != "false"]`, just `["${BUILD_PDF:-}" != "false"]`. Better yet, the usual convention is to test `[ "${BUILD_PDF:-}" ]` which will let a user set `BUILD_PDF` to anything to mean true and empty or unset to mean false. Using a command-line argument is a more expected interface in most cases though.

You may also find the [Hoffman Lab application command-line user-interface checklist](https://docs.google.com/document/d/1xJSXHjkgcl77K_-d14YpC5xqY3epQwDYCS_eEYgq6dg) helpful.

dhimmel:
Thanks @michaelmhoffman for the excellent feedback and shell wisdom.

I'll work on integrating your points. In the longterm, it is not clear that we will continue managing the build with a shell script, rather than say migrating the pipeline into Python. So given that consideration, I will probably avoid large changes like adding command-line arguments to the shell scripts... until at least we have a better idea of the ideal future architecture. However, many other suggestions are immediately actionable and I'm excited to implement them!

build/build.sh
```diff
@@ -56,7 +56,7 @@ pandoc --verbose \
 DOCKER_EXISTS="$(command -v docker || true)"
 
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
-if [ "${BUILD_PDF:-true}" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
+if [ "$(BUILD_PDF:-true)" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
```
michaelmhoffman: This should be parameter expansion (`${BUILD_PDF:-true}`) not command substitution (`$(command -v docker)`). Same below for some variables.

dhimmel: Ah that distinction was lost on me, and the root cause of all this back and forth incorrect usage!

build/build.sh
```diff
@@ -1,12 +1,18 @@
-set -o errexit
+#!/usr/bin/env bash
+
+## build.sh: compile manuscript outputs from content using Manubot and Pandoc
+
+set -o errexit \
+    -o nounset \
+    -o xtrace
```
agitter: The suggestion to not set `xtrace` here and instead allow it optionally with `bash -x script.sh` makes sense to me.

ci/deploy.sh
```diff
@@ -1,16 +1,22 @@
-# Exit on errors
-set -o errexit
+#!/usr/bin/env bash
+
+## deploy.sh: run during a Travis CI build to deploy manuscript outputs to the output and gh-pages branches on GitHub.
+
+# Set options for extra caution & debugging
+set -o errexit \
+    -o nounset \
+    -o xtrace
```
agitter: Same `xtrace` comment as above

build/build.sh
```diff
@@ -1,12 +1,18 @@
...
 
 # Set timezone used by Python for setting the manuscript's date
 export TZ=Etc/UTC
 # Default Python to read/write text files using UTF-8 encoding
 export LC_ALL=en_US.UTF-8
 
 # Generate reference information
-echo "Retrieving and processing reference metadata"
+>&2 echo "Retrieving and processing reference metadata"
```
agitter: I'm not accustomed to seeing `>&2` at the beginning of the line, but it looks like there is not single convention for this.

build/build.sh
```diff
@@ -54,8 +60,8 @@ pandoc --verbose \
 DOCKER_EXISTS="$(command -v docker || true)"
 
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
-if [ "$BUILD_PDF" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
-  echo "Exporting PDF manuscript using WeasyPrint"
+if [ "${BUILD_PDF:-true}" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
```
agitter: Change to use michaelmhoffman's tip 11? https://github.com/manubot/rootstock/pull/233#issuecomment-500898800

build/build.sh
michaelmhoffman: I agree this seems rather odd to me but I did not comment on it earlier. More important is a consistent use of whatever convention you choose. My own preference in descending order would be:

```
echo >&2 "Retrieving and processing reference metadata"
echo "Retrieving and processing reference metadata" >&2
>&2 echo "Retrieving and processing reference metadata"
```

Normally I do redirections at the end of a command but with `echo` or `printf` it can get lost at the end.

build/build.sh
dhimmel: Changed in https://github.com/manubot/rootstock/pull/233/commits/da0b397784d27d259e532b6d79c35405f56673b6 along with adding `-o pipefail`. From these [GNU docs](https://www.gnu.org/software/bash/manual/html_node/Pipelines.html)

> The exit status of a pipeline is the exit status of the last command in the pipeline, unless the pipefail option is enabled (see The Set Builtin). If pipefail is enabled, the pipeline’s return status is the value of the last (rightmost) command to exit with a non-zero status, or zero if all commands exit successfully. If the reserved word ‘!’ precedes the pipeline, the exit status is the logical negation of the exit status as described above. The shell waits for all commands in the pipeline to terminate before returning a value.

So I am not sure why `pipefail` does not interfere with the error suppression in:

```
DOCKER_EXISTS="$(command -v docker || true)"
```

build/build.sh
dhimmel: > You don't need `["${BUILD_PDF:-true}" != "false"]`, just `["${BUILD_PDF:-}" != "false"]`. Better yet, the usual convention is to test `[ "${BUILD_PDF:-}" ]` which will let a user set `BUILD_PDF` to anything to mean true and empty or unset to mean false. Using a command-line argument is a more expected interface in most cases though.

I don't think `[ "${BUILD_PDF:-}" ]` is what we wan't here because it defaults to not building the PDF if `BUILD_PDF` is unset. However, we want PDF building unless explicitly disabled with `BUILD_PDF=false`. I also think we want a solution where `BUILD_PDF=false` and `BUILD_PDF=true` perform as written, such that they can be easily toggled in CI configs. We could do `["${BUILD_PDF:-}" != "false"]`, which saves a few characters but is a bit less readable for the non-expert.

michaelmhoffman: You could call it `SUPPRESS_PDF` or better yet `MANUBOT_SUPPRESS_PDF` (or something semantically equivalent).

Another alternative is simply to set the variable upstream with something like:

```
BUILD_PDF="${BUILD_PDF:-true}"
```

Then you can just use `"${BUILD_PDF}"` from then on forward.

build/build.sh
michaelmhoffman: `command -v docker || true` is a list of commands, not a pipeline.

dhimmel:
Ready for review. I addressed several of @michaelmhoffman's suggestions, but not all of them. Some I'd like to address at a future time, but don't want to hold up this PR. Would be good for reviewers to take one more pass to make sure I haven't done anything incorrect.

dhimmel:
> I would advise adding shellcheck to your editor and your CI workflow which would catch some of the errors above.

For the record, I satisfied all shellcheck grievances besides quoting:

https://github.com/manubot/rootstock/blob/2fd84f1484858245acbdc8acf37167a378aeeca4/ci/deploy.sh#L19-L20

Did not modify those lines to avoid conflicts since manuscript instances change them.

> You may want some of these variables to be set only in cases where the user hasn't set and exported them.

I started modifying this, but stopped because I remembered we had problems in the past and would want to make sure we're not introducing a bug regression. Also it's not entirely clear to me that we shouldn't always use UTC timezone for manuscripts. Anyways, just want to spend some more time learning how these variables and configurations work before blindly changing.

agitter:
That reasoning make sense to me.

Your comment about regression testing is what I was hinting at.  Because the `manubot` package and the Rootstock content are still evolving, we can't easily test that changes to the build process still generate identical manuscripts.  However, is there a lighter form of testing we may eventually want?

The `manubot` package is already testing the `process` command so we don't need to test that here.

dhimmel:
> However, is there a lighter form of testing we may eventually want?

I can think of two possibilities:

1. Move more rootstock functionality to the manubot package where we have a testing framework. build.sh would probably be the hardest to migrate, but parts of deploy.sh could be part of the python library. Although testing deployment is also super difficult.

2. Have a manuscript repository with extra tests that somehow updates whenever rootstock does. Or potentially switch from a clone based setup method to a template based on, as per https://github.com/manubot/rootstock/issues/6#issuecomment-316452685. Then perhaps we could have extra tests in the template repo.

No immediate easy solutions come to mind.

PR #234, opened 2019-06-25
Add links to PLOS Comp Bio article on Manubot

dhimmel:
Also update URLs in `02.delete-me.md`


dhimmel:
> Or did you want to keep that pointed to the Manubot version?

I was thinking we'd keep that with the Manubot version, which now links to the PLOS version for DOI-citing. In general, I think the Manubot version is better for reading (especially for users interested in seeing how Manubot works). The PLOS version is better for citing.

PR #238, opened 2019-06-28
Improve documentation about build process

### LINKED ISSUE #237: Local-path figures not rendered in output/manuscript.html
Hi,
I am learning manubot and the learning curve is very fast! I really like it! Thanks for creating it!  
Two quick questions:  
- My figures are not rendered in .html, but they are rendered in .pdf. Do you have any suggestions on what I should double-check?  
- When I run `build.sh` I get a bunch of warnings (see attachment). Do you have any recommendations on how to fix them?
Thanks a lot!
Serena 

warnings

agitter:
Some of the documentation about the local build script had not been updated for a while.  @sbonaretti would these updates help address the undocumented behavior that led to your questions in #237?

Closes #237

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ The directories are as follows:
 
 ### Local execution
 
-The easiest way to use Manubot is to use [continuous integration](#continuous-integration) to rebuild the manuscript when the content changes.
+The easiest way to run Manubot is to use [continuous integration](#continuous-integration) to rebuild the manuscript when the content changes.
 If you want to build a Manubot manuscript locally, install the [conda](https://conda.io) environment as described in [`build`](build).
 Then, you can build the manuscript on POSIX systems by running the following commands from this root directory.
 
@@ -58,7 +58,7 @@ bash build/build.sh
 # Configure the webpage directory
 python build/webpage.py
 
-# View the manuscript locally at http://localhost:8000/
+# View the manuscript webpage/index.html locally at http://localhost:8000/
 cd webpage
 python -m http.server
 ```
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ The directories are as follows:
 
 + [`content`](content) contains the manuscript source, which includes markdown files as well as inputs for citations and references.
   See [`USAGE.md`](USAGE.md) for more information.
-+ [`output`](output) contains the outputs (generated files) from the manubot including the resulting manuscripts.
++ [`output`](output) contains the outputs (generated files) from Manubot including the resulting manuscripts.
   You should not edit these files manually, because they will get overwritten.
 + [`webpage`](webpage) is a directory meant to be rendered as a static webpage for viewing the HTML manuscript.
 + [`build`](build) contains commands and tools for building the manuscript.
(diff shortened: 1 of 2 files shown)


sbonaretti:
It looks good to me! Thanks for changing and for asking! :)

build/README.md
```diff
@@ -17,7 +17,8 @@ Please reference [Pull Request #40](https://github.com/manubot/rootstock/pull/40
 
 Note: currently, **Windows is not supported**.
 
-Install or update the [conda](https://conda.io) environment specified in [`environment.yml`](environment.yml) by running:
+Install or update the [conda](https://conda.io) environment specified in [`environment.yml`](environment.yml) by running
+the following commands from this directory:
```
dhimmel: probably don't want this newline

build/README.md
```diff
@@ -33,3 +34,16 @@ However, it will fail on Windows due to the [`pango`](https://anaconda.org/conda
...
 Because the build process is dependent on having the appropriate version of the `manubot` Python package, it is necessary to use the version specified in `environment.yml`.
 The latest `manubot` release on PyPI may not be compatible with the latest version of this rootstock repository.
+
+## Building PDFs
+
+If Docker is available, `build.sh` uses the [Athena](https://www.athenapdf.com/) [Docker image](https://hub.docker.com/r/arachnysdocker/athenapdf) to build the PDF.
+Otherwise, `build.sh` uses [WeasyPrint](https://weasyprint.org/) to build the PDF.
+It is common for WeasyPrint to generate many warnings and errors that can be safely ignored.
+Examples are shown below:
+```sh
```
dhimmel: I'd remove `sh`... this is not shell, just stdout to the shell.

README.md
```diff
@@ -52,11 +53,12 @@ bash build/build.sh
 
 # At this point, the HTML & PDF outputs will have been created. The remaining
 # commands are for serving the webpage to view the HTML manuscript locally.
+# This is required to view local images in the HTML output.
 
 # Configure the webpage directory
 python build/webpage.py
 
-# View the manuscript locally at http://localhost:8000/
+# View the manuscript webpage/index.html locally at http://localhost:8000/
```
dhimmel: We should probably note that usually you should be able to just open `webpage/index.html` in a browser and it will work. However, to open a local webserver use the following:


This has changed from before. The HTML outputs now work better when just openning in browser without web server.

build/README.md
```diff
@@ -33,3 +33,16 @@ However, it will fail on Windows due to the [`pango`](https://anaconda.org/conda
...
+If Docker is available, `build.sh` uses the [Athena](https://www.athenapdf.com/) [Docker image](https://hub.docker.com/r/arachnysdocker/athenapdf) to build the PDF.
+Otherwise, `build.sh` uses [WeasyPrint](https://weasyprint.org/) to build the PDF.
+It is common for WeasyPrint to generate many warnings and errors that can be safely ignored.
+Examples are shown below:
+```
+WARNING: Ignored `pointer-events: none` at 3:16, unknown property.
+WARNING: Ignored `font-display:auto` at 1:53114, descriptor not supported.
+ERROR: Failed to load font at "https://use.fontawesome.com/releases/v5.7.2/webfonts/fa-brands-400.eot#iefix"
+WARNING: Expected a media type, got only/**/screen
+```
```
dhimmel: Perhaps we should say something like:

> Setting `BUILD_PDF=false` will disable all PDF creation.

agitter: At the top of the file we have
> However, setting the `BUILD_PDF` environment variable to `false` will suppress PDF output.
For example, run local builds using the command `BUILD_PDF=false bash build/build.sh`.

Should we move that or leave it there?

dhimmel: Ah I missed that from the diff. Your call.

agitter: I'll leave it for now. Feel free to move it later if we make another pass through the documentation.

PR #245, opened 2019-07-04
Add customization instructions and catalog to USAGE

### LINKED ISSUE #242: How to include another CSL?
I wasn't able to find in the documents how I can implement another CSL file to render my references based on some specific requirements.

I think modifying the line `CSL_PATH=build/assets/style.csl` in `build.sh` is the way. Is there an easier way without changing `build.sh`?

agitter:
Closes #242 

This may end up being a placeholder until we have more permanent documentation.

Should we remove the list of example manuscripts here and focus on maintaining the catalog?

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -192,7 +192,7 @@ If you experience any issues with the Manubot or would like to contribute to its
 
 ## Examples
 
-For additional examples, check out existing manuscripts that use the Manubot (some of which are still in progress):
+For additional examples, check out existing manuscripts that use Manubot in the [Manubot catalog](https://manubot.org/catalog/) or the list below:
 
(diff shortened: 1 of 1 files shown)


dhimmel:
> Should we remove the list of example manuscripts here and focus on maintaining the catalog?

Yes! This way never a great place for them. Perhaps we could remove the current list and refer to the catalog instead. We should also probably reference try-manubot (somewhere in USAGE).

agitter:
I added the catalog and try-manubot links at the top of the page.  I also confirmed that all of the examples there were here are in the catalog, except https://slochower.github.io/synthetic-motor-literature/.

slochower:
@agitter Ah, right. Unfortunately that project is on hiatus, so that manuscript won't be updated anytime soon, but it was a good early introduction to Manubot!

agitter:
@slochower is to okay to drop it as an example for now and have you add it back at https://github.com/manubot/catalog if you update it?  It's not a problem having an unfinished manuscript in the examples, but it looks like `manubot cite` is unable to identify you as the author.

It is still listed as an existing manuscript in the Manubot paper.

slochower:
@agitter yep!

dhimmel:
> it looks like manubot cite is unable to identify you as the author

Yeah old Manubot versions didn't set the author metadata in a way translation-server can detect. I forget exactly what changed, but it may make sense to not include the mansucript in catalog until it's updated.

Issue #240, opened 2019-07-01
merging adjacent table cells


sbonaretti:
Hi! Is there a way to merge to (horizontally) adjacent cells in a table?
Thanks!!!

slochower:
I don't think this is possible until the upstream Markdown processor that `pandoc` uses, implements it. There is a very long thread discussing the possibility and implementation details: https://github.com/jgm/pandoc/issues/1024. I can't quite tell what the timescale is for this, but I'm confident that a lot of people care about this and the authors of `pandoc` seem to be taking it seriously.

vincerubinetti:
Unfortunately the only way to accomplish this right now is to write out the table "manually" (in html) in your markdown, and use `colspan`, like this:

```HTML
<table>
 <tr>
  <td colspan="2">&nbsp;</td>
  <td>&nbsp;</td>
 </tr>
 <tr>
  <td>&nbsp;</td>
  <td>&nbsp;</td>
  <td>&nbsp;</td>
 </tr>
 <tr>
  <td>&nbsp;</td>
  <td>&nbsp;</td>
  <td>&nbsp;</td>
 </tr>
</table>
```
![image](https://user-images.githubusercontent.com/8326331/60444507-62e05680-9beb-11e9-93b5-c8eeff9fe3b0.png)


Reference:
https://www.computerhope.com/issues/ch001655.htm

dhimmel:
> Unfortunately the only way to accomplish this right now is to write out the table "manually" (in html) in your markdown

@vincerubinetti are you sure the HTML table method works when passing through Pandoc? My understanding is that since Pandoc's Astract Syntax Tree doesn't support colspans, if the table gets interpreted by Pandoc, this solution won't work. However, I'm not sure whether Pandoc would interpret the table or just pass-through the HTML verbatim.

vincerubinetti:
Just tested it, it does work. My understanding is that anything you write in plain HTML within a markdown file will just get passed through/ignored by pandoc. Is that consistent with your understanding?

slochower:
It will pass through to the HTML output, but did you check that it also passes through to PDF and DOCX? That part I'm not sure about.

vincerubinetti:
PDF output comes right from the HTML output, so as long as WeasyPrint/Athena supports `colspan`, it should work. As for `docx`, I don't know but I doubt it.

slochower:
I think that writing the table as HTML will also break the functionality used by by `pandoc-tablenos`.

[...]

PR #251, opened 2019-07-09
add attributes plugin

vincerubinetti:
In response to and closes https://github.com/manubot/rootstock/issues/240

This PR adds a plugin the allows the user to add arbitrary HTML attributes -- including inline styles -- to almost any HTML element. Simply write an HTML comment in the following format next to or inside of the desired element:

`<!-- $attribute="value" -->` (or `<!-- $attribute='value' -->`)

I wrote this in such a way that the user writes it exactly how they'd write it in HTML. This plugin is for more advanced use, and somewhat requires some knowledge of HTML and what are valid attributes for which types of elements.

@dhimmel @agitter Test this out. I've tested it on almost everything in the `delete-me.md` and it only doesn't work for codeblocks. You can attach it to paragraphs, links, spans, quotes, table cells, and more.

Also, note, when I say "next to", you don't necessarily have to type out the element in HTML. It can be in markdown, like so:

```
| Tables                      |     Are     |  Cool |
|-----------------------------|:-----------:|------:|
| col 1 <!-- $colspan="2" --> |             | $1600 |
| col 2                       |     money   |   $12 |
| col 3                       |     money   |    $1 |
```


slochower:
Neat! Can we test an example with spanning columns like #240?

vincerubinetti:
@slochower Yes, that works. See edit to the original issue comment.

build/plugins/attributes.html
```diff
@@ -0,0 +1,132 @@
...
+                    options[optionName] = param;
+            }
+        }
+        loadOptions();
+
+        // start script when document is finished loading
+        if (options.enabled === 'true')
+            window.addEventListener('load', start);
+    })();
+</script>
```
dhimmel: newline

build/plugins/attributes.html
```diff
@@ -0,0 +1,132 @@
...
+                const regex2 = /([a-zA-Z\-]+)?=\"(.+?)\"/;
+                const regex1 = /([a-zA-Z\-]+)?=\'(.+?)\'/;
+
+                // loop through attribute/value pairs
+                let match;
+                let count = 0;
+                while(
+                    match = text.match(regex2) || text.match(regex1),
+                    count++,
+                    match && count < 100
```
dhimmel: What is the significance of 100 here?

build/build.sh
```diff
@@ -48,6 +48,7 @@ pandoc --verbose \
   --include-after-body=build/plugins/link-highlight.html \
   --include-after-body=build/plugins/table-of-contents.html \
   --include-after-body=build/plugins/lightbox.html \
+  --include-after-body=build/plugins/attributes.html \
```
dhimmel: Perhaps we want to keep this disabled by default until we have some more real world testing and we're sure the API is stable?

build/plugins/attributes.html
```diff
@@ -0,0 +1,132 @@
...
+
+<script>
+    (function() {
+        // /////////////////////////
+        // DESCRIPTION
+        // /////////////////////////
+
+        // This Manubot plugin allows arbitrary HTML attributes to be attached
+        // to (almost) any element. Place an HTML comment inside or next to the
+        // desired element in the format <!-- attribute="value" -->
```
dhimmel: Can the comment elaborate on the behavior of "next to the desired" element? Also should we describe the colspan special case behavior?

build/plugins/attributes.html
```diff
@@ -0,0 +1,132 @@
...
+                if (!comment.parentElement)
+                    continue;
+
+                // get text content of comment
+                let text = comment.nodeValue.trim();
+
+                // regex's for finding attribute/value pairs in the format of
+                // attribute="value" or attribute='value
+                const regex2 = /([a-zA-Z\-]+)?=\"(.+?)\"/;
+                const regex1 = /([a-zA-Z\-]+)?=\'(.+?)\'/;
```
dhimmel: I see `-` is allowed in attribute but not `_`. Do HTML attributes not use underscores?

Does this only match from the start of the comment string? For example what happens if:

```markdown
<!-- The previous sentence is needed to describe when method="funk" -->
```

Probably should make syntax isn't detected by accident (although if it were, the effects would probably be minor?).

The initial comment uses `<!-- colspan: 2 -->`, but that syntax is different than `<!-- colspan="2" -->`

build/plugins/attributes.html
vincerubinetti: This is an arbitrary safety limit so we don't crash the tab if anything accidentally goes wrong in the while loop. I don't know all the possible edge cases of what the javascript `match` function can return, so it's slightly possible that the "remove from string" step might fail in a way that causes an infinite loop.

However, there's while loops in the other plugins, and I don't do this sort of safety limit. Those I feel a little more confident about (that they're predictable), but perhaps I should consider putting safety cutoffs in those cases as well.

build/plugins/attributes.html
vincerubinetti: I don't know how to elaborate on that in a helpful way, because it varies for every element.

If it's a link, you put it in between the square brackets. If it's a header, put it right after the text on the same line. If it's a paragraph, put it at the end of any sentence.

It depends on how Pandoc passes the comment through and what element it ends up attaching the comment to, and that decision is somewhat arbitrary, dependent on pandoc, not documented by pandoc (I assume), and might also be subject to change in the future.

build/plugins/attributes.html
vincerubinetti: Standard HTML attributes only have letters (`colspan` `style` `class`) or dashes (`data-somevar` `stroke-width`). Custom attributes can have more, but custom attributes won't do anything without associated javascript, so at that point the person has to be a javascript developer anyway.

See this for more info: https://stackoverflow.com/questions/925994/what-characters-are-allowed-in-an-html-attribute-name

Also no, it doesn't need to be at the start of the string. That is intentional. And it matches multiple, so you can put more than one attribute (it'd be frustrating to the user for us to make this plugin but then only allow one attribute). Also, I think other notes should be allowed right next to the attributes.

Therefore I think the behavior is correct as it is. `<!-- this is some extra styling style="color: red" and here's a colspan colspan="2" yay -->`

vincerubinetti:
> Where in the script is this exclusion handled?

This is actually handled implicitly. My code specifically loops through `COMMENT` type nodes. When the browser loads the page, it builds an element/node representation of the structure of the document, aka the DOM. If it sees the syntax `<!-- -->` it creates a `COMMENT` node there.

When pandoc sees that same syntax within a piece of code, it escapes the necessary characters such that the browser sees it as a plain text node instead of a comment node. And thus, the plugin wont even pick it up.

build/plugins/attributes.html
dhimmel: I agree that defining multiple attributes in a single comment is important.

I don't like however that the plugin scans every comment and extracts things where the user may not actually be intending to set attributes. In your example, the user does intent to set attributes. In my example, the user didn't. Therefore, would it make sense to require something at the beginning of the comment like:

```
<!-- htmlattrs colspan="2" style="color: red" -->
```

Then you could strip leading whitespace from all comments and ignore any comment that doesn't begin with `htmlattrs` (or whatever keyword we choose).

Finally, it seems that the `colspan="2" style="color: red"` definitions are adopting a standard syntax for HTML? If so, would it be possible to use the builtin infrastructure to parse this text? This way it could deal with more complex cases (possibly with escaping etc) and would work exactly as if this text had been actual HTML.

One final consideration is how the system will work for non-English users. It seems like HTML tags are always in English, such that this is not a problem?

build/plugins/attributes.html
dhimmel: Ah I see. The attributes apply to whatever element pandoc places them in. The "next to the desired element" confused me. Let's tweak the documentation to elaborate on this point.

build/plugins/attributes.html
vincerubinetti: I don't know of any existing infrastructure to do this. You're not really intended to do anything like this in normal practice. There is `setAttribute` and `getAttribute` for setting/getting attributes, but that is javascript.

As far as I know, yes, the attribute names are always in english. Attributes that accept strings, like `title` or `alt`, I believe can accept any characters/languages.

Regarding `htmlattrs`, we can talk more about how you'd like to set this up such that users only do it intentionally.

dhimmel: Is this the right functionality? https://api.jquery.com/jQuery.parseHTML/

vincerubinetti: Not really, no. That jQuery function parses a full html string, like `<span class="button">text</span>`, and returns a DOM node that can be inserted into the document.

In our case, we already have the DOM node (the node that the comment is attached to/under) and just need to add attributes to it.

What you're looking for (I think) is an existing function to read a string and extract out valid HTML attribute/value pairs. I don't believe that exists.

As for your second links, they do give me an idea:

I can duplicate the parent node, find the end of the opening tag, and insert the comment string into it like `<span this text was originally in a comment colspan="2">`, read that duplicate's attributes with `getAttribute`, and apply the same values to the real parent node.

But I think this is more error prone, more verbose, still requires some parsing (to find the starting tag and add text to it), and still not a standard procedure.

dhimmel: If we dissallow non-attribute comments after colspan, would this be straightword:

From here:

```html
<!-- htmlattrs colspan="2" style="color: red" -->
```

To here:

```
<span colspan="2" style="color: red" ></span>
```

To parsing the attributes and extracting them?

vincerubinetti: That particular change doesn't help or hurt.

I can rewrite the plugin using the idea in my previous comment, and you can see how you like the readability of it compared to the current method. The new method is less predictable, to me. I can't be sure there won't be unforeseen odd edge cases that cause trouble down the road.

vincerubinetti: After experimentation, I've concluded that the "new" method can't actually work because cloning the node adequately is not possible. You can verify this for yourself by running these snippets in the console:

Try to clone object by first creating container then setting `innerHTML` with new desired attribute string from comment
```javascript
container = document.createElement("div");
container.innerHTML = "<td newAttribute='fromComment'></td>";
console.log(div);
```
The resulting `<div>` node contains nothing because a `<td>` (table cell) is not valid unless it's a child of a `<tr>` (row) and `<table>`.

This problem also occurs using `DOMParser`. There are certain combinations and fragments of HTML that are not allowed, and it would be too difficult to capture them all and handle the exceptions. It would almost be like rewriting an HTML display engine.

Also, for the situations where the above approach (and the several variants that I also tried) DOES work, it still doesn't seem to "purify" the input in any useful way. It still returns all of the attributes in a list, even if they're not valid HTML attributes. It seems to just split the string by spaces and `=`'s. Try out this code:

```javascript
div = document.createElement("div");
div.innerHTML = "<span text: text colspace='2'>text</span>";
console.log(div.firstChild.attributes);
```

Therefore I don't think this is better in any way than the current approach.

vincerubinetti:
@dhimmel What about prepending a single symbol, like `<!-- comment text #colspan="2">`. Or maybe `!` or `%` or `$` (like jQuery).

If not a symbol, I'd vote for `htmlattr `, with a single space, no more no less, between it and the attribute name.

vincerubinetti:
Reminder for me to create an issue to discuss the possible syntaxes for this.

Probably will be Monday or later.

vincerubinetti:
@dhimmel this is ready to be merged after you review my changes to the delete-me. I give an example of colspan, and a more general example.

content/02.delete-me.md
```diff
@@ -217,6 +226,15 @@ _Journals are not likely to support them, and they may not display correctly whe
...
 [Link styled as a button](https://manubot.org "Manubot Homepage"){.button}
 
+Adding arbitrary HTML attributes with the `attributes` plugin:
+
+Manubot Manubot Manubot Manubot Manubot.
+Manubot Manubot Manubot Manubot.
+Manubot Manubot Manubot.
+Manubot Manubot.
+Manubot.
+<!-- $id="paragraph_of_text" $style="color: #ad1457; margin-left: 40px;" $disabled="true" $title="a paragraph of text" $data-color="red" -->
```
dhimmel: In this case, pandoc syntax could also be used to set these values, so should we add a note that says something like "for demonstration, since these attributions can also be set using pandoc extensions or html syntax"?

vincerubinetti: I’d rather also have a paragrah right above this one demonstrating the pandoc syntax

vincerubinetti: I’d rather also have a paragrah right above this one demonstrating the pandoc syntax

build/build.sh
```diff
@@ -40,14 +40,14 @@ pandoc --verbose \
   --csl="$CSL_PATH" \
   --metadata link-citations=true \
   --include-after-body=build/themes/default.html \
-  --include-after-body=build/plugins/table-scroll.html \
```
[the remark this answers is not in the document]
vincerubinetti: No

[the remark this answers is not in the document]
vincerubinetti: No

vincerubinetti:
@dhimmel this is ready for re-review and merge

content/02.delete-me.md
```diff
@@ -192,14 +192,23 @@ White background specified to serve as a backdrop for transparent sections of th
...
-| pi | 3.14159265358979323846264338327950 | 288419716939937510582097494459230 | 781640628620899862803482534211706 | [`piday.org`](https://www.piday.org/million/) |
-| e  | 2.71828182845904523536028747135266 | 249775724709369995957496696762772 | 407663035354759457138217852516642 | [`nasa.gov`](https://apod.nasa.gov/htmltest/gifcity/e.2mil) |
+|         | Digits 1-33                        | Digits 34-66                      | Digits 67-99                      | Ref.                                                        |
+|:--------|:-----------------------------------|:----------------------------------|:----------------------------------|:------------------------------------------------------------|
+| pi      | 3.14159265358979323846264338327950 | 288419716939937510582097494459230 | 781640628620899862803482534211706 | [`piday.org`](https://www.piday.org/million/)               |
+| e       | 2.71828182845904523536028747135266 | 249775724709369995957496696762772 | 407663035354759457138217852516642 | [`nasa.gov`](https://apod.nasa.gov/htmltest/gifcity/e.2mil) |
```
dhimmel: Do you think all this white-space padding is helpful? Oftentimes, I think wide markdown tables are best in compact format (one space like `| pi | 3.14159265358979323846264338327950 |`). The cells are not aligned in a text editor, but since most text editors wrap long lines, they won't be aligned anyways.

Would it make sense to keep non-compact spacing in table below, but switch to compact in the very wide table here?

vincerubinetti: I just thought it looked ugly personally. I usually view my files unwrapped, so it helps me.

I made sure there was an example of a table without exact spacing, so they know that it's not a requirement.

Issue #252, opened 2019-07-10
Doubling of equations in athena PDF output


berceanu:
When comparing the [HTML output](https://berceanu.github.io/lwfa_scaling/) with the [PDF output](https://berceanu.github.io/lwfa_scaling/manuscript.pdf) of my manuscript, I noticed all the (non-inline) formulas appear twice in the `.pdf`. Here is the original [repo](https://github.com/berceanu/lwfa_scaling).

agitter:
This could possibly be related to using Athena to build the PDF.  When I locally built this manuscript during the BOM testing I used weasyprint.  The equations weren't duplicated:
[manuscript.pdf](https://github.com/manubot/rootstock/files/3377827/manuscript.pdf)

It may also be relevant that the two copies of the equations use different fonts.

dhimmel:
[Here's a permalink](https://berceanu.github.io/lwfa_scaling/v/6cbf02dc6d1c6cbba15b49d6b418823357fd9c34/manuscript.pdf) to the PDF with the duplicated equations. One interesting thing is that the duplications are not present on the final few pages.

Here is our current athenapdf command:

https://github.com/manubot/rootstock/blob/2a5e425650947020d2d4b49447e3f3d9718dee40/build/build.sh#L90-L98

I wonder if increasing the delay from 2 seconds would help? Perhaps the page is still rendering. When I [print the HTML to PDF](https://berceanu.github.io/lwfa_scaling/v/6cbf02dc6d1c6cbba15b49d6b418823357fd9c34/) in Chrome, I don't get duplications. This should be a similar process as what Athena is doing (albeit with different versions of dependencies). @berceanu has this happened multiple times? If you rerun CI will that fix it? In other words, is this an intermittent error.

@vincerubinetti any ideas?

> When I locally built this manuscript during the BOM testing I used weasyprint. The equations weren't duplicated

@agitter I thought that equations didn't render with WeasyPrint because they require javascript. That is bizarre

berceanu:
Might also be solved if https://github.com/manubot/rootstock/issues/249 is addressed.

vincerubinetti:
I think I have a good idea of what is causing this. These equations are being rendered with MathJax right?

[...]

PR #255, opened 2019-07-22
disable mathjax fast preview

vincerubinetti:
This PR disables MathJax's fast-preview option and increases the Athena print delay in response to #252.

Here's what it looks like now to load equations:

record

You can briefly see some gray text, which I believe is just the plain text of the latex input. It's then replaced by an SVG when it's ready.

It looks like there is a brief moment where the both the gray text and the svg is there. So I guess even with this change, it's possible that equations could still be doubled if the timer for pdf output is not long enough.

http://docs.mathjax.org/en/latest/options/extensions/fast-preview.html

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -60,6 +60,7 @@ pandoc --verbose \
 DOCKER_EXISTS="$(command -v docker || true)"
 
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
+# If Docker is not available, use WeasyPrint to create PDF
 if [ "${BUILD_PDF:-}" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
   echo >&2 "Exporting PDF manuscript using WeasyPrint"
   if [ -L images ]; then rm images; fi  # if images is a symlink, remove it
@@ -82,9 +83,12 @@ if [ "${BUILD_PDF:-}" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
   rm images
 fi
 
-# Create PDF output (unless BUILD_PDF environment variable equals "false")
+# If Docker is available, use athenapdf to create PDF
 if [ "${BUILD_PDF:-}" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
   echo >&2 "Exporting PDF manuscript using Docker + Athena"
+  if [ "${CI:-}" = "true" ]; then
+    echo >&2 "Continuous integration build detected. Setting athenapdf --delay=${MANUBOT_ATHENAPDF_DELAY:=5000}"
+  fi
   if [ -d output/images ]; then rm -rf output/images; fi  # if images is a directory, remove it
   cp -R -L content/images output/
   docker run \
@@ -94,7 +98,7 @@ if [ "${BUILD_PDF:-}" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
     --security-opt=seccomp:unconfined \
     arachnysdocker/athenapdf:2.16.0 \
     athenapdf \
-    --delay=5000 \
+    --delay=${MANUBOT_ATHENAPDF_DELAY:-1100} \
     manuscript.html manuscript.pdf
   rm -rf output/images
 fi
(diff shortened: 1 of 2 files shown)


dhimmel:
I think the following implementation of conditionally setting `MANUBOT_ATHENAPDF_DELAY` makes sense:

```sh
# before athenapdf command
if [ "${CI:-}" = "true" ]; then
  echo "continuous integration build detected. Setting athenapdf --delay=${MANUBOT_ATHENAPDF_DELAY:=5000}"
fi

# for athenapdf --delay
--delay=${MANUBOT_ATHENAPDF_DELAY:-1100}
```

https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

dhimmel:
@vincerubinetti and @agitter does https://github.com/manubot/rootstock/pull/255/commits/b45fabc204cfc09c0f6561ecc2a6aeea05176156 look good to you? I tested the bash locally and it seems to work as intended. I wanted to keep the athenapdf command working even for users that copy and paste it and lose the context of the build.sh script.

dhimmel:
@michaelmhoffman if you have time to look at the shell script changes in this PR that would be appreciated... it's only a few lines.

build/build.sh
```diff
@@ -94,7 +98,7 @@ if [ "${BUILD_PDF:-}" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
     --security-opt=seccomp:unconfined \
     arachnysdocker/athenapdf:2.16.0 \
     athenapdf \
-    --delay=2000 \
+    --delay=${MANUBOT_ATHENAPDF_DELAY:-1100} \
```
agitter: What does the `-` indicate?

build/build.sh
```diff
@@ -82,9 +83,12 @@ if [ "${BUILD_PDF:-}" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
   rm images
 fi
 
-# Create PDF output (unless BUILD_PDF environment variable equals "false")
+# If Docker is available, use athenapdf to create PDF
 if [ "${BUILD_PDF:-}" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
   echo >&2 "Exporting PDF manuscript using Docker + Athena"
+  if [ "${CI:-}" = "true" ]; then
```
agitter: Should we add a comment here explaining why the delay is longer when running on CI?  It is a little counter intuitive.  One might expect that the delay should always be long enough to avoid mathjax problem regardless of whether we build locally or remotely.

That said, the decision does make sense to me.  We are less sensitive to build time when running on Travis CI.  We also want the "official" versions of the manuscripts hosted on GitHub pages to be less likely to contain errors.

build/build.sh
dhimmel: From https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

> `${parameter:-word}`:
If parameter is unset or null, the expansion of word is substituted. Otherwise, the value of parameter is substituted.

So `:-` gets interpreted as a single operator. This PR also uses `:=`:

> `${parameter:=word}`
If parameter is unset or null, the expansion of word is assigned to parameter. The value of parameter is then substituted. Positional parameters and special parameters may not be assigned to in this way.

build/build.sh
dhimmel: https://github.com/manubot/rootstock/pull/255/commits/8119a99b5317d169907b2aecb5c40ce3366c24a7

build/build.sh
agitter: Answering my own question with https://unix.stackexchange.com/questions/122845/using-a-b-for-variable-assignment-in-scripts

build/build.sh
```diff
@@ -82,9 +83,14 @@ if [ "${BUILD_PDF:-}" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
...
+    # Incease --delay for CI builds to ensure the webpage fully renders, even when the CI server is under high load.
+    # Local builds default to a shorter --delay to minimize runtime, assuming proper rendering is less crucial.
+    echo >&2 "Continuous integration build detected. Setting athenapdf --delay=${MANUBOT_ATHENAPDF_DELAY:=5000}"
```
michaelmhoffman: Setting the default value of a variable as a side-effect of printing a log message does not lead to understandable code.

dhimmel: Not sure I totally prefer https://github.com/manubot/rootstock/pull/255/commits/44576670255d8a8482bdc6bdddb120a898396bce, but if it will be more widely understood, then OKAY with me.

PR #259, opened 2019-08-02
Disable Jekyll processing on GitHub Pages

dhimmel:
Specify `--no-jekyll` to add a `.nojekyll` file to the `gh-pages` branch.

Manubot does not make use of Jekyll to do additional processing of webpage content hosted on GitHub Pages. Disable Jekyll to bypass an unnecessary CI job, which can slow down webpage deployment. https://github.blog/2009-12-29-bypassing-jekyll-on-github-pages/

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -84,6 +84,7 @@ ghp-import \
 
 # Deploy the webpage directory to gh-pages
 ghp-import \
+  --no-jekyll \
   --follow-links \
   --push \
   --branch=gh-pages \


dhimmel:
The other option would be:

```
touch webpage/.nojekyll
```

But I find the proposed solution cleaner because it avoids cluttering master with an additional file and keeps the `webpage` directory host agnostic.

agitter:
This looks okay to me, but I'm not familiar with Jekyll processing.  Should we have @vincerubinetti review before merging?

vincerubinetti:
This is fine with me. It shouldn't affect anything.

dhimmel:
> I'm not familiar with Jekyll processing

GitHub Pages automatically performs Jekyll processing. It's like an extra CI step that runs on `gh-pages`, but where you can't see the CI logs. In our case, it doesn't modify the content of `gh-pages` at all.

Currently, gh-pages --> jekyll --> GitHub Pages
After this PR, gh-pages --> GitHub Pages

Issue #112, opened 2018-02-15
Support Alternate Themes


Miserlou:
It is difficult to read a long manuscript with the current style settings.

It might be useful to build on the work of other projects which convert Markdown into the usual academic style:

https://github.com/ickc/markdown-latex-css
https://github.com/thomaspark/pubcss/ // https://thomaspark.co/project/pubcss/demo/acm-sig-sample-web-theme.html
https://gist.github.com/killercup/5917178
etc

dhimmel:
> It is difficult to read a long manuscript with the current style settings.

What exactly do you dislike?

> It might be useful to build on the work of other projects which convert Markdown into the usual academic style

Our current [`github-pandoc.css`](https://github.com/greenelab/manubot-rootstock/blob/310dd078c4c60bedd63442f93906f26a4ea25a4b/webpage/github-pandoc.css) was based off of [this gist](https://gist.github.com/dashed/6714393).

Regarding the alternatives:

+ Markdown Latex CSS [looks](https://ickc.github.io/markdown-latex-css/) pretty basic and outdated IMO
+ [PubCSS](https://github.com/thomaspark/pubcss/) seems promising. It [looks like](https://thomaspark.co/project/pubcss/demo/acm-sig-sample-web-theme.html) it can produce nice HTML pages that when printed take on more a traditional journal article appearance.
+ Regarding https://gist.github.com/killercup/5917178, I think I may have tested it. Is likely similar to our current `github-pandoc.css`. Would need to see evidence that it's better.

Anyways, the frontend stuff is a bit outside my expertise, so we'd need a contributor to take the lead if we wanted alternative viewers. Another longterm option would be to use a JATS XML viewer, such as [Lens Viewer](https://github.com/elifesciences/lens) (dependent on JATS export in https://github.com/greenelab/manubot-rootstock/pull/82). Then the webpage could look [like this](https://lens.elifesciences.org/about/).

Note that currently the way we generate the PDF is to use WeasyPrint to essentially print the webpage to PDF. Ideally, we want nice HTML and PDF output.

dhimmel:
If you're looking for the most beefy Manubot manuscripts out there, the following are good options:

- https://github.com/greenelab/scihub-manuscript
- https://github.com/greenelab/deep-review
- https://github.com/dhimmel/rephetio-manuscript (outdated Manubot version)
- https://github.com/dhimmel/bitcoin-whitepaper/ (contains math)

[...]

PR #261, opened 2019-08-02
[stub] support alternate HTML theme through environment variable

michaelmhoffman:
Change to `build.sh` demonstrating potential solution to #112.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -23,6 +23,7 @@ manubot process \
 CSL_PATH=build/assets/style.csl
 BIBLIOGRAPHY_PATH=output/references.json
 INPUT_PATH=output/manuscript.md
+THEME_HTML_PATH="${THEME_HTML_PATH:-build/themes/default.html}"
 
 # Make output directory
 mkdir -p output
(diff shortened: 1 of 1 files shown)


agitter:
I like this design.  We should consider doing the same for `CSL_PATH`, `BIBLIOGRAPHY_PATH`, and `INPUT_PATH`.

dhimmel:
Noting that if we adopt `pandoc --defaults` as proposed in https://github.com/manubot/rootstock/pull/290, this method of setting pandoc options through environment variables will no longer work.

michaelmhoffman:
```
PANDOC_DEFAULTS_PATH="${PANDOC_DEFAULTS_PATH:-build/pandoc-defaults.yaml}"
```

dhimmel:
We merged https://github.com/manubot/rootstock/pull/290, which includes something similar to @michaelmhoffman's suggestion:

https://github.com/manubot/rootstock/blob/06fb01d0c8e5c1ca7632cd17c27d3fddd0691386/build/build.sh#L22-L24

Now that we're using `--defaults` files, this PR is less relevant, so am closing it.

To customize pandoc's behavior, you can either edit the relevant line a default YAML file, or create a new pandoc-defaults directory and change `PANDOC_DEFAULTS_DIR` as discussed at https://github.com/manubot/rootstock/pull/290#discussion_r362842345.

Issue #198, opened 2019-03-23
Exposing rendered manuscripts from pull request builds


dhimmel:
Currently, rendered manuscripts created by pull request builds, such as `output/manuscript.html` and `output/manuscript.pdf`, are never exposed. Instead they disappear with the ephemeral build environment.

Exposing rendered manuscripts (and ideally a rendered diff, see #54) would make review of pull requests easier. For example, currently we don't know how a citation will render until we merge the PR (or test it locally). Nor do we know if formatting syntax will work as intended. The point has also been brought up by a reviewer of the manubot manuscript in https://github.com/greenelab/meta-review/issues/121.

The main issue blocking us is that pull requests on Travis cannot access secure environment variables due to their unauthenticated nature. This limitation is discussed at https://github.com/travis-ci/travis-ci/issues/5579. Possibly there are workarounds or alternatives to using PR build artifacts.

agitter:
To learn more about our options, should we test an alternative CI service?  CircleCI supports [artifacts](https://circleci.com/docs/2.0/artifacts/).  That wouldn't provide the complete solution we are looking for. It may provide a quick short term workaround to the specific problem of reviewing rendered manuscripts for a PR.

dhimmel:
I would be interested to see what an operational `.circleci/config.yml` looks like, so we can evaluate how it compares to Travis. However, switching CI providers would be highly disruptive, so I'm cautious to have us invest too much time in trying out Travis alternatives, until we have a roadmap that accommodates CI migration. Certain designs may also allow us to support multiple CI providers.

agitter:
I tested building a manuscript with AppVeyor, which supports build artifacts in pull requests, to assess how much that would help.  https://github.com/agitter/my-manuscript/pull/4 provides a rough demonstration.

The overall strategy is to continue using Travis CI for deployment and use a parallel build on AppVeyor to store a copy of the updated `output/manuscript.pdf` from the pull request.  The build process in the AppVeyor [config file](https://github.com/agitter/my-manuscript/blob/e62c70fbd8cbeee487a488270582a0ea004c32a2/.appveyor.yml) is fairly similar to the `.travis.yml` file.  The AppVeyor [build](https://ci.appveyor.com/project/agitter/my-manuscript/builds/26413750#L367) writes the URL of the latest PDF:
```
Updated PDF available from https://ci.appveyor.com/api/projects/agitter/my-manuscript/artifacts/output/manuscript.pdf?branch=add-ref
```
[...]

PR #262, opened 2019-08-03
Add AppVeyor pull request previews

agitter:
Partially addresses #198.  Tested using https://github.com/agitter/my-manuscript/.  See https://github.com/agitter/my-manuscript/pull/9#issuecomment-517947114 for an example pull request notification.

I still need to get the shared Travis CI and AppVeyor setup to work from a script.

@dhimmel do I have permissions to enable AppVeyor on this repository?  Or can you please do it?


agitter:
The Travis CI build now works with an install script.  Once we enable AppVeyor for this repository and test that build, this is ready for review.

.appveyor.yml
```diff
@@ -0,0 +1,46 @@
+# See https://www.appveyor.com/docs/getting-started-with-appveyor-for-linux/
+skip_branch_with_pr: true
+
+# AppVeyor will run on the output branch commits until the commit message
```
dhimmel: Do you mean "unless" rather than "until"? Or are you saying CI runs on the initial empty commit like https://github.com/manubot/rootstock/commit/13115af95ca1b4daadd5242a4f3285557f8b87a8? It seems like our initial commits [do include](https://github.com/manubot/rootstock/blob/3defccfbea9d088f3b6885e1bf18158c6a05f467/SETUP.md#create-repository) `[ci skip]`:

```
git commit --allow-empty \
  --message "Initialize empty branch" \
  --message "[ci skip]"
```

.appveyor.yml
```diff
@@ -0,0 +1,46 @@
...
+# contains [ci skip]
+# Enable 'Do not build on "Push" events' in the AppVeyor project settings
+# to only build commits from pull requests
+branches:
+  only:
+    - master
+
+only_commits:
+  files:
+    - content/
```
dhimmel: This is cool but perhaps too restrictive. For example, don't we want to test that changes to environment.yml or .appveyor.yml don't break AppVeyor builds?

.appveyor.yml
```diff
@@ -0,0 +1,46 @@
...
+
+image: ubuntu
+services:
+  - docker
+
+install:
+  - source ci/install.sh
```
dhimmel: Ah I wonder if we `source` rather than bash on travis, we could get each command to be in the log?

ci/deploy.sh
```diff
@@ -59,7 +59,7 @@ fi
 
 # Commit message
 MESSAGE="\
-$(git log --max-count=1 --format='%s')
+$(git log --max-count=1 --format='%s') [ci skip]
```
dhimmel: This has to be in the title of the commit message to work?

dhimmel:
# Activating AppVeyor for this repository

1. Go to https://ci.appveyor.com/projects/new
2. Click "update installations"
3. for "Where do you want to install AppVeyor?" chose "configure" the manubot GitHub organization. If we didn't already have the organization linked, I think it would say "Request"
4. Only select repositories... choose `manubot/rootstock`
5. Click on `+Add` after hovering over `manubot/rootstock`.
6. CI is activated at https://ci.appveyor.com/project/manubot/rootstock. Note that I believe `manubot` in this URL does not refer to the github manubot organization, but instead our account on appveyor.

agitter:
> do you have control over the CI?

I believe I have access to the Manubot team now.  I can access its AppVeyor account but don't see myself listed as a member at https://ci.appveyor.com/account/manubot/team
image
I sent myself another invite using my email address associated with my AppVeyor account, and it said I had already joined the team.  Seems like it worked.

> Note that I believe manubot in this URL does not refer to the github manubot organization, but instead our account on appveyor.

That's right, if I recall correctly from past projects with my `gitter-lab` organization.

.appveyor.yml
agitter: I meant "unless".  Fixed and expanded in 0817820

.appveyor.yml
agitter: Yes, I wanted to demonstrate the possibility but am undecided on the best behavior.  I want to avoid notification fatigue.  I'm also assuming we'll rely on Travis CI builds for most of our traditional continuous integration and use AppVeyor only for pull request review.

An intermediate option would be to add more build-related files to this whitelist.  We should be able to run AppVeyor if the content, `environment.yml`, `.appveyor.yml`, or any `.sh` files in the repository change.

ci/deploy.sh
agitter: Yes, AppVeyor requires it to be in the first line of the commit message.  Check my build history https://ci.appveyor.com/project/agitter/my-manuscript/history

Build `1.0.64` ran on the `output` branch commit https://github.com/agitter/my-manuscript/commit/08eef2c19a79f3f6c6ec7ced196d9e1d9b891264 even though it contains `[ci skip]` in the middle of the commit message.  It didn't run on https://github.com/agitter/my-manuscript/commit/dbadde6859c2b23c0758e3e3fde6ecfca4063c2a once I moved `[ci skip]` to the first line.

agitter:
I made a test commit (now reverted) that changed the content to trigger an AppVeyor build.  That worked as expected.

We still need to decide whether to build all commits, only those that affect whitelisted files, or all commits except those that only affect blacklisted files (e.g. README.md).

dhimmel:
> We still need to decide whether to build all commits, only those that affect whitelisted files, or all commits except those that only affect blacklisted files (e.g. README.md).

Seems like whitelisting the following directories would be sufficient: `content`, `build`, `ci`. Also do we need to whitelist `.appveyor.yml` or is that assumed?

I'll get started on the manubot CI environment detection implementation.

RHagenson:
If I can add my two-cents as an outsider and fan of this project, setting up AppVeyor previews appears to be intended to build a preview of pull requests prior to acceptance where after the preview effectively becomes the content on gh-pages. Am I mistaken on this point? 

Building on all PRs would then maintain consistency in the PR review process, as well as eliminate all possible mistakes in whitelist/blacklist-ing files properly as a project expands. In my opinion, the one submitting a PR should not need to know what is whitelisted or blacklisted in order for their PR to be reviewed the same as any other.

dhimmel:
Hey @RHagenson, thanks for your input. Your understanding is correct, the AppVeyor previews allow viewing the PDFs that commits in a pull request create before having merged the PR. 

> Building on all PRs would then maintain consistency in the PR review process, as well as eliminate all possible mistakes in whitelist/blacklist-ing files properly as a project expands.

Yes, however it could lead to cluttering PRs with AppVeyor comments and notification fatigue. For example, if someone creates a PR to update a README, then every commit in the PR will trigger a comment like https://github.com/manubot/rootstock/pull/262#issuecomment-518757386. IMO it'd be best to avoid these AppVeyor comments unless there has been an actual change. An alternative solution would be for the user to put `[skip ci]` in their commit message, although it's easy to forget this.

I think that the whitelist approach is pretty safe in terms of not building commits that actually effect the manuscript, but let's see what @agitter thinks is most prudent.

agitter:
Thanks @RHagenson, outside opinions are always welcome.  We want to make sure Manubot is useful to more people than the core team.

>  setting up AppVeyor previews appears to be intended to build a preview of pull requests prior to acceptance

Yes, that's right.  Ideally we could have done this within the Travis CI builds, but Travis CI does not support persistent artifacts like AppVeyor, Circle CI, etc.  The point is indeed to help with pull request review.  Currently, it is hard for a reviewer to know whether a reference's metadata will be extracted as expected unless they manually check via the Manubot command line interface or something similar.

I see the case for building on all commits to keep things simple.  The only real downside would be excessive notifications in projects that frequently modify files that do not affect the manuscript.  For instance, https://github.com/Benjamin-Lee/deep-rules often has pull requests to change the [contributor tracking table](https://github.com/Benjamin-Lee/deep-rules/blob/master/contributors.md) that do not affect the manuscript or build process.

dhimmel:
@agitter I committed https://github.com/manubot/rootstock/pull/262/commits/4aec0e2c6362ae05ff0bfc11128bc18daf5d90cd to output the list of environment variables for help with https://github.com/manubot/manubot/pull/131. AppVeyor did not run even though `.appveyor.yml` was edited, so looks like we have it explicitly whitelist it as I do in https://github.com/manubot/rootstock/pull/262/commits/1e8d10df75e0185b3dbfab6a1bcf8095980f8bcd.

As per [these docs](https://www.appveyor.com/docs/how-to/filtering-commits/), we could also add:

```
only_commits:
  message: /\[force ci\]/
```

This would allow users to force the appveyor build via commit message in case their whitelisting rules are too stringent.

RHagenson:
I understand the desire to avoid notification fatigue. In that case, I think the whitelist approach is best with the addition of clearly documenting that only edits under specific directories (e.g., `content`, `build`, and `ci`, as previously mentioned) trigger previews. 

Not too familiar with the manubot/rootstock build process (been following it, but not yet used it) so as long as there is no way to leak content into other directories that would then need whitelisting and complicate the build process I see no reason why whitelisting could not be used. Ultimately, my opinion on this is driven by this repo being the initial template others use so any decision forces a process upon potential users  -- the less complex the forced process, the more likely users are to continue developing with manubot.

.appveyor.yml
```diff
@@ -0,0 +1,48 @@
...
+  - env
+    
+test_script:
+  - bash build/build.sh
+  - MANUSCRIPT_FILENAME=manuscript-$APPVEYOR_BUILD_VERSION-${APPVEYOR_REPO_COMMIT:0:7}.pdf
```
dhimmel: Because pull request builds make a merge commit rather than just build the commit from the PR, the commit shown in MANUSCRIPT_FILENAME is currently the merged commit rather than the commit that shows up in the PR. Perhaps it would be better for user recognition to use something like:

```bash
TRIGGERING_COMMIT = ${APPVEYOR_PULL_REQUEST_HEAD_COMMIT:-APPVEYOR_REPO_COMMIT}
MANUSCRIPT_FILENAME=manuscript-$APPVEYOR_BUILD_VERSION-${TRIGGERING_COMMIT:0:7}.pdf
```

.appveyor.yml
```diff
@@ -0,0 +1,48 @@
...
+  - echo "Updated PDF available from https://ci.appveyor.com/api/buildjobs/$APPVEYOR_JOB_ID/artifacts/$MANUSCRIPT_FILENAME"
+
+# The following lines can be safely deleted, which will disable AppVeyorBot
+# notifications in GitHub pull requests
+# Notifications use Mustache templates http://mustache.github.io/mustache.5.html
+# See https://www.appveyor.com/docs/notifications/#customizing-message-template
+# for available variables
+notifications:
+  - provider: GitHubPullRequest
+    template: "{{#passed}}:white_check_mark:{{/passed}}{{#failed}}:x:{{/failed}} [Build {{&projectName}} {{buildVersion}} {{status}}]({{buildUrl}}) (commit {{commitUrl}} by @{{&commitAuthorUsername}})
```
dhimmel: can we get this to show the triggering commit?

.appveyor.yml
```diff
@@ -0,0 +1,48 @@
...
+  - cp output/manuscript.pdf $MANUSCRIPT_FILENAME
+  - appveyor PushArtifact $MANUSCRIPT_FILENAME
+
+build: off
+
+cache:
+  - ci/cache
+
+on_success:
+  - echo "Artifacts available from https://ci.appveyor.com/project/$APPVEYOR_REPO_NAME/builds/$APPVEYOR_BUILD_ID/artifacts"
```
dhimmel: Can replace `https://ci.appveyor.com` with `$APPVEYOR_URL`.

`$APPVEYOR_REPO_NAME` should be `$APPVEYOR_ACCOUNT_NAME/$APPVEYOR_PROJECT_NAME` because here we need the AppVeyor account info rather than the GitHub account info.

.appveyor.yml
agitter: The triggering commit isn't directly accessible through the webhook payload, but I found a workaround.  AppVeyor suggests using messages to transfer custom information from the build process into the payload.  We can create a message with the triggering commit and then use that message in the pull request notification.

The end result is a notification like https://github.com/agitter/my-manuscript/pull/11#issuecomment-518842928

What do you think about that?  A side effect is a message in the build that doesn't mean much on its own https://ci.appveyor.com/project/agitter/my-manuscript/builds/26515001/messages

dhimmel: > What do you think about that? A side effect is a message in the build that doesn't mean much on its own

Potentially we can move more of the github comment text to the message to make it more understandable in the build context. Let me propose an ideal notification text.

.appveyor.yml
agitter: Done in bcd55532168cbe77ab4e0c9453539ae8b6f88aea

.appveyor.yml
agitter: Done in bcd55532168cbe77ab4e0c9453539ae8b6f88aea

I like this better as long as we also use the triggering commit in the notification message below.

.appveyor.yml
agitter: That would help.  However, I didn't find the equivalent of the webhook's `commitAuthorUsername` in the environment variables, and I like reporting that username instead of the `APPVEYOR_REPO_COMMIT_AUTHOR`.

dhimmel: One downside of these PR comments is that they will end up triggering a bunch of email notifications, which I don't think are useful (since the commit email notification implies AppVeyor will start). However, doing an `@` mention probably won't increase the number of emails in most cases, because the user mentioned is likely already subscribed to the PR. So OKAY with me.

SETUP.md
```diff
@@ -196,3 +196,14 @@ When there are no longer any unstaged changes, then do `git commit`.
...
+# Previewing pull request builds with AppVeyor
+
+You can optionally enable AppVeyor continuous integration to view pull request builds.
+Unlike Travis CI, AppVeyor supports storing manuscripts generated during pull request builds as artifacts.
+These can be previewed to facilitate pull request review and ensure formatting and reference changes render as expected.
+When a pull request build runs successfully, the AppVeyorBot will post a notification in the pull request linking to the versioned manuscript PDF.
+
+To enable AppVeyor, follow steps 1 and 2 of the [AppVeyor welcome](https://www.appveyor.com/docs/) to sign in to AppVeyor and add your manuscript repository as an AppVeyor project.
+The repository already contains an `.appveyor.yml` build configuration file, so no other setup is required.
+AppVeyor is configured to only run when the files in the `content` directory are modified.
```
dhimmel:
```suggestion
AppVeyor only runs when it detects changes that are likely to effect the manuscript.
```

.appveyor.yml
agitter: I didn't think it is necessary to run AppVeyor if `ci/deploy.sh` changes because that is only used by Travis CI.  I also want to use the `.appveyor.yml` to explain or document these whitelisted files so that we don't have to maintain a separate list in `SETUP.md`.

ci/deploy.sh
dhimmel: > Yes, AppVeyor requires it to be in the first line of the commit message.

Looked into this a bit more. [They require it](https://help.appveyor.com/discussions/problems/543-skip-ci-not-working#comment_44000370) to be before any blank lines in the commit message. A misguided implementation IMHO.

agitter: If AppVeyor looks for blank lines, should we change this to
```
$(git log --max-count=1 --format='%s')
[ci skip]
```

dhimmel: We can do

```
$(git log --max-count=1 --format='%s')
[ci skip]
```

So that [ci skip] is on the second line. Not really standard git practice but perhaps better than lengthening the title.

dhimmel:
Would the following be a better notification comment:

> The AppVeyor [build](https://ci.appveyor.com/project/manubot/rootstock/builds/26529062) for commit  54c33fb by @agitter is now complete. The rendered manuscript from this build is temporarily available for download at [`manuscript-1.0.12-54c33fb.pdf`](https://ci.appveyor.com/api/buildjobs/2yexi9lxa9xm6jdi/artifacts/manuscript-1.0.12-54c33fb.pdf).

dhimmel:
We probably want to include https://github.com/manubot/manubot/pull/131 before merging this PR? I think I am going to update 131 so it  can create the permalinks.

agitter:
I like your notification message.  We could make a minor tweak to include the build number in the message and show what that number in the filename refers to

> AppVeyor [build 1.0.12](https://ci.appveyor.com/project/manubot/rootstock/builds/26529062) for commit  54c33fb by @agitter is now complete. The rendered manuscript from this build is temporarily available for download at [`manuscript-1.0.12-54c33fb.pdf`](https://ci.appveyor.com/api/buildjobs/2yexi9lxa9xm6jdi/artifacts/manuscript-1.0.12-54c33fb.pdf).

For a failed build we could use
> AppVeyor [build 1.0.12](https://ci.appveyor.com/project/manubot/rootstock/builds/26529062) for commit  54c33fb by @agitter failed.

I'll need to test a failed build.  We may need to create the AppVeyor message earlier in case the build fails during the install step.

 > We probably want to include manubot/manubot#131 before merging this PR?

Because there isn't any urgency, I agree.  That way we can update the environment in this same pull request.  I'll review that other PR soon but won't get to it today.

dhimmel:
> For a failed build we could use

I'm not sure we need a comment at all if there's a failure... since the GitHub UI will show the failed test. The comment doesn't provide anything that isn't already readily available.

agitter:
I'd prefer to only notify on success.  The problem is that this discussion indicates AppVeyor doesn't support conditional notifications https://help.appveyor.com/discussions/questions/10376-only-send-a-notification-based-on-some-condition

My interpretation of that is that some notification will be triggered for every build and we can only control what the message is conditionally.  Unless we construct our own workaround notifications.

dhimmel:
Sometimes a build may fail but a PDF gets created nonetheless. It can be helpful to see the PDF to diagnose the build failure sometimes (like if its citation related). Would there be any way to link to the manuscript artifact for failures if it exists?

Also I am okay with continuing to refine the AppVeyor setup in subsequent PRs, if we'd like to get this merged sooner.

agitter:
> Would there be any way to link to the manuscript artifact for failures if it exists?

Yes, that should be possible.  I'm testing builds under a few different conditions so we can see if the notifications work as expected under these different scenarios.

agitter:
@dhimmel these notification messages are good enough for now.  We can see if we want to revise them once we start using them in real workflows.

I'd like to review once more end to end before merging to make sure I didn't leave in any test code, but I'm not planning any more updates.  I'm also okay waiting for https://github.com/manubot/manubot/pull/131.

agitter:
This is ready for final review and merge.  I suggest we make a separate pull request to follow up on https://github.com/manubot/manubot/pull/131 that updates the front matter and Manubot version.

When we merge, I'd also like to remove the URL in the commit message of 70c86f27f6a547fc2a51ac8a6dc0fc5dcfc4e5c1.  It's no longer relevant so we don't need to spam that linked issue.

PR #263, opened 2019-08-09
Use AppVeyor artifact URLs in manuscript front-matter

dhimmel:
Follows up on https://github.com/manubot/rootstock/pull/262#issuecomment-519944443
Refs https://github.com/manubot/rootstock/issues/198

### DIFF
--- a/content/00.front-matter.md
+++ b/content/00.front-matter.md
@@ -11,7 +11,7 @@ _A DOI-citable version of this manuscript is available at <https://doi.org/DOI_H
 {## Template to insert build date and source ##}
 <small><em>
 This manuscript
-{% if ci_source is defined -%}
+{% if ci_source is defined and ci_source.manuscript_url is defined -%}
 ([permalink]({{ci_source.manuscript_url}}))
 {% endif -%}
 was automatically generated
--- a/content/00.front-matter.md
+++ b/content/00.front-matter.md
@@ -12,7 +12,7 @@ _A DOI-citable version of this manuscript is available at <https://doi.org/DOI_H
 <small><em>
 This manuscript
 {% if ci_source is defined -%}
-([permalink](https://{{ci_source.repo_owner}}.github.io/{{ci_source.repo_name}}/v/{{ci_source.commit}}/))
+([permalink]({{ci_source.manuscript_url}}))
 {% endif -%}
 was automatically generated
 {% if ci_source is defined -%}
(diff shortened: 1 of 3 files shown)


dhimmel:
The updated conda version to have shaved aboutr 32 seconds off build time in terms of environment installation.

content/00.front-matter.md
```diff
@@ -12,7 +12,7 @@ _A DOI-citable version of this manuscript is available at <https://doi.org/DOI_H
 <small><em>
 This manuscript
 {% if ci_source is defined -%}
```
agitter: Because `ci_source.manuscript_url` is created in a separate Manubot function `add_manuscript_urls_to_ci_params`, should we also ensure that `ci_source.manuscript_url`  is defined here?  Or only check for `ci_source.manuscript_url`?

content/00.front-matter.md
```diff
@@ -12,7 +12,7 @@ _A DOI-citable version of this manuscript is available at <https://doi.org/DOI_H
...
-([permalink](https://{{ci_source.repo_owner}}.github.io/{{ci_source.repo_name}}/v/{{ci_source.commit}}/))
+([permalink]({{ci_source.manuscript_url}}))
 {% endif -%}
 was automatically generated
 {% if ci_source is defined -%}
```
agitter: It's not obvious to me which commit we should show in line 19 below for AppVeyor builds.  The manuscript filename contains the triggering commit, but this link contains the merge commit for a pull request build.  However, the triggering commit doesn't correspond to the `ci_source.repo_slug` tree.  It resides in the tree of the forked repository.

I don't see a better option that what is already here.  I do expect the difference between the filename and link could cause confusion.

dhimmel: Yes it is confusing, but I think it's important that the manuscript contains the commit that actually created it.

content/00.front-matter.md
dhimmel: I don't think `ci_params` will ever exist without `manuscript_url` with the current manubot software because:

https://github.com/manubot/manubot/blob/1b885c614634d6e88565ebcf5052b6884132ab32/manubot/process/util.py#L193-L196

But it doesn't hurt to be safe and possibly some users will have an outdated `manubot` python package with this version of `content/00.front-matter.md `, so I added the extra check in https://github.com/manubot/rootstock/pull/263/commits/b07d870959f583f0fb6424dd1648d3795411999b

Issue #224, opened 2019-06-03
Links don't wrap correctly


Benjamin-Lee:
In the rendered HTML, wrapping does not preserve words inside of links. For example:

<img width="755" alt="image" src="https://user-images.githubusercontent.com/13618205/58773408-f55cee00-858a-11e9-9657-74c25c7c42ac.png">

agitter:
Cross referencing related discussion in https://github.com/greenelab/deep-review/pull/951

dhimmel:
@vincerubinetti any solutions you can think of that would be better than the current behavior, which causes this odd wrapping within short words / citation numbers?

vincerubinetti:
@Benjamin-Lee Can you provide a link to the html output for convenience, if possible? 

You're not talking about the pdf export by saying "rendering HTML" are you?

dhimmel:
Quoting from https://github.com/greenelab/deep-review/pull/951#issuecomment-498260768

> To clarify on this, there is no CSS way to apply a certain style to an element based on the text it contains. In other words, if you want to wrap based on the length of the link text, or try to wrap if the link text appears to be a URL and not just a sentence, that will require javascript.

Would it make sense to default to no special wrapping rules for links, but then using JS to find long hyperlinks whose text is URL-like and then applying the wrap option to just those elements?

dhimmel:
Note that we could also consider a class like `.wrap` that users could apply to problematic long URLs. Also use table context to wrap hyperlinked text in cells?

dhimmel:
Talking with @vincerubinetti, the `word-break: break-all` that could be causing this may be fixed in the latest rootstock. Now it only applies to to table cells:

https://github.com/manubot/rootstock/blob/091ca8d85c8ef2d7af16fcc8d2ed3ebcbc187f13/build/themes/default.html#L224-L232

vincerubinetti:
> Note that we could also consider a class like `.wrap` that users could apply to problematic long URLs. Also use table context to wrap hyperlinked text in cells?

A [`<wbr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr) might be sufficient. Haven't used them before, so will have to look into their effectiveness.

dhimmel:
The document where this is happening is https://benjamin-lee.github.io/deep-rules/v/e80c29cf64bbc0018edc9161e2483b33bea59225/#fig:overfitting-fig

vincerubinetti:
[...]

PR #266, opened 2019-09-05
fix link overflow

vincerubinetti:
This has been bugging me for a while:

image

Long links that can't break at a "natural" language point -- like links that have a lot of numbers I guess, it's dependent on the browser -- cause overflow to the right of the body/paper element.

I remember that I had previously had `word-break: break-all` for links, but that was causing them to break at unnatural points, like in between words, so we took it out, not remembering why it was there in the first place.

This PR adds it back in, but with `break-word`, which forces line breaks for long number links, but should also try to break them at natural points first.

___

Just to be sure, tested that it breaks at words:

image

image

### DIFF
--- a/build/themes/default.html
+++ b/build/themes/default.html
@@ -102,6 +102,7 @@
     /* links */
     a {
         color: #2196f3;
+        word-break: break-word;
     }
 
     /* normal links (not empty, not button link, not syntax highlighting link) */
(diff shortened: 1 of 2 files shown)


vincerubinetti:
@dhimmel need review

dhimmel:
@vincerubinetti did you confirm this change doesn't reintroduce the issues in https://github.com/manubot/rootstock/issues/224 and https://github.com/greenelab/deep-review/pull/951#issuecomment-475671180 where short words were breaking.

vincerubinetti:
@dhimmel yes, see the latter two screenshots in the original post

vincerubinetti:
https://www.nikpro.com.au/difference-between-overflow-wrap-and-word-break-with-examples/

Issue #270, opened 2019-10-09
Printing with Chrome on a Mac / default settings -> words run off page on US letter


cgreene:
Printing the PDF generated at https://greenelab.github.io/biopriors-review/v/7e8a3a2c7f58bf8fee3c1547870f54f98141fdb1/manuscript.pdf from Chrome on a mac  leads to text that runs off the page.

> ![Screen Shot 2019-10-09 at 2 12 15 PM](https://user-images.githubusercontent.com/542643/66508411-0f5d8e80-ea9f-11e9-9319-b430ffec21bb.png)

This isn't apparent before the print dialog in chrome:

> ![Screen Shot 2019-10-09 at 2 11 59 PM](https://user-images.githubusercontent.com/542643/66508424-184e6000-ea9f-11e9-8ca6-2af4e2ea1a57.png)

dhimmel:
The PDF uses the paper size of "A4, Portrait (8.26 × 11.69 inch)". On a sheet of paper with this size, the margins will appear correctly. The issue is that our lab printer in the U.S. uses "letter" paper  (8.50 × 11.00 inch). [This page](http://betweenborders.com/wordsmithing/a4-vs-us-letter/) visualizes the size difference:

> ![image](https://user-images.githubusercontent.com/1117703/66510265-ae37ba00-eaa2-11e9-9727-fed66ce020f6.png)

If we switch to letter, then A4 paper will have the sides cutoff.

I can imagine a couple options:

1. Instruct users to print using "fit to size" scaling to reduce the dimensions of the text if needed.
2. Make sure our PDF margins accommodate paper with dimensions of 8.26 × 11.00 inch (the min in both dimensions).

***

We've previously struggled with this issue in https://github.com/manubot/rootstock/issues/105 and https://github.com/manubot/rootstock/pull/107. In https://github.com/manubot/rootstock/commit/a535f7b872df775b1557f90b01402bbd56d17517, we switched to letter size, but this is no longer part of the default CSS:

https://github.com/manubot/rootstock/blob/631b8a2d9efe8fc8c7b0fe8bf7d4cfe5239bcfb0/build/themes/default.html#L485-L489

So I guess athenapdf is defaulting to A4.

vincerubinetti:
I think it makes more sense to specify a default size in whatever is actually doing the printing, rather than in the CSS which is usually harder to change. Thus I don't think we should hard code it into the CSS.

Two additional half measures we discussed in person:

[...]

PR #272, opened 2019-10-11
Explicitly specify PDF pagesize

dhimmel:
Closes https://github.com/manubot/rootstock/issues/270

The existence of two common pagesizes for physical paper, A4 and Letter, creates user confusion. Sometimes users print a PDF with A4 dimensions onto Letter paper and do not use the "scale to fit" printer setting. This leads to truncated text at the bottom of the page. Ultimately, it is not a Manubot issue, but confuses Manubot users nonetheless.

When exporting the HTML to PDF Manubot uses an A4 pagesize, which while less common in the U.S. is an international standard. This commit explicitly specifies the default behavior, such that there is no functional change. However, by specifying the pagesize, its more clear to Manubot users where pagesize gets set. This will hopefully spare users extra debugging.

If manuscript viewers would like to export to a pagesize that is different from the one used by manuscript.pdf, the recommendation is for them to print the HTML page to PDF from their browser and specify the desired page size there. For this reason, our CSS style does not specify a default page size, but instead just a margin.

From [arachnys/athenapdf source](https://github.com/arachnys/athenapdf/blob/bf3d5d0070b079bc38f76f56aeb2ca09f3b3454c/cli/src/athenapdf.js#L39):

> athenapdf --pagesize <size>
> page size of the generated PDF (default: A4, values: A3|A4|A5|Legal|Letter|Tabloid)

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -103,6 +103,7 @@ if [ "${BUILD_PDF:-}" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
     arachnysdocker/athenapdf:2.16.0 \
     athenapdf \
     --delay=${MANUBOT_ATHENAPDF_DELAY:-1100} \
+    --pagesize=A4 \
     manuscript.html manuscript.pdf
   rm -rf output/images
 fi


dhimmel:
We can't do the same thing for WeasyPrint [because](https://weasyprint.readthedocs.io/en/stable/tutorial.html):

> Currently, WeasyPrint does not provide support for adjusting page size or document margins via command-line flags. This is best accomplished with the CSS `@page` at-rule.

Not sure what happens by default (whether A4 is always used or its locale specific).

PR #274, opened 2019-10-31
Update environment on 2019-10-31

dhimmel:


dhimmel:
https://github.com/manubot/rootstock/pull/274/commits/fa83dc1f3c442dea553f68d51db1828f65c56f7c updates usage to mention the `url` workaround for https://github.com/manubot/manubot/issues/16 and https://github.com/manubot/manubot/issues/158. CC @jmonlong. I don't want to close those issues until they are fixed upstream by bioRxiv / Crossref. But this should be of some help perhaps.

Issue #288, opened 2019-12-17
Improved documentation


Nebucatnetzer:
Are there any ideas floating around regarding an improved documentation?

dhimmel:
Right now documentation is scattered between READMEs, USAGE.md, SETUP.md, `content/02.delete-me.md` and the python package README and [autogenerated docs](https://manubot.github.io/manubot/). Consolidating these docs to whatever extent makes sense is something we think about. We're also looking to simplify how Manubot works and how many files it requires... so this could simplify the documentation task.

Are there any holes in the documentation that you'd like to let us know about?

vincerubinetti:
In short, I just want to say: I'm on it.  As Daniel said, we're in the process of simplifying Manubot itself, so it made sense to sort of hold off on the full documentation until that was complete, because it will change a lot.

Nebucatnetzer:
To give some context, I'm currently trying to recreate my thesis (written in latex) with manubot to have some sort of real life example.
I just noticed that I'm hopping around a lot to find stuff (from USAGE.md to issues, to pandoc, to common mark, etc).
One example was adding page breaks which I didn't found in the USAGE.md but found an issue (#35) about it and the option was actually quite nice.
Other things I thought about and didn't find in USAGE.md were: mathjax (found it in build.sh), toc, footnotes, glossary.

However I'm unsure about what you want to include in Manubot since I've read somewhere that it is more intended to be for reviews which might exclude a thesis like mine (it's not an academic paper).

As for the shifting nature of the project. How about something simple, like the repo wiki here on Github?
It's not pretty and IMO a bit limited if the project grows to much but written in Markdown and easy to use.
The auto generated docs look great of course.

vincerubinetti:
:thumbsup: definitely. A road map would be a good thing to state upfront clearly too, so people don't have to go through hundreds of github issues to see if we're even going to attempt doing a certain feature.

Nebucatnetzer:
I see we're on the same page here :), for the road map maybe the Kanban board might be useful?
If you already know some things which you won't include it might be good to create issues and close them right away with "wontfix".

Should I add things like the page break to the USAGE.md file to begin with so that they are a bit easier to find and don't get lost until a proper solution was found for the documentation?

vincerubinetti:
I think we'd all welcome any PR you'd like to contribute in the meantime!

dhimmel:
[...]

PR #289, opened 2019-12-18
Update metadata.authors & USAGE

dhimmel:
metadata.yaml should now use `authors` instead of `author_info`.

Update USAGE to mention delete-me.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -12,6 +12,9 @@ Markdown files are identified by their `.md` extension and ordered according to 
 For basic formatting, check out the [CommonMark Help](https://commonmark.org/help/) page for an introduction to the formatting options provided by standard markdown.
 In addition, Manubot supports an extended version of markdown, tailored for scholarly writing, which includes [Pandoc's Markdown](https://pandoc.org/MANUAL.html#pandocs-markdown) and the extensions discussed below.
 
+The `content/02.delete-me.md` file in the Rootstock repository shows many of the elements and formatting options supported by Manubot.
+See the [raw markdown](https://gitlab.com/manubot/rootstock/blob/master/content/02.delete-me.md#L) in this file and compare it to the [rendered manuscript](https://manubot.github.io/rootstock/).
+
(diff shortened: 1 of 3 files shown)


USAGE.md
```diff
@@ -179,7 +182,7 @@ The following YAML shows the supported key–value pairs for an author:
 ```yaml
 github: dhimmel  # strongly suggested
 name: Daniel S. Himmelstein  # mandatory
-initials: DSH  # strongly suggested
+initials: DSH  # optional
```
dhimmel: initials are not used anywhere by default, so seems excessive to call them "strongly suggested"

USAGE.md
```diff
@@ -12,6 +12,9 @@ Markdown files are identified by their `.md` extension and ordered according to
 For basic formatting, check out the [CommonMark Help](https://commonmark.org/help/) page for an introduction to the formatting options provided by standard markdown.
 In addition, Manubot supports an extended version of markdown, tailored for scholarly writing, which includes [Pandoc's Markdown](https://pandoc.org/MANUAL.html#pandocs-markdown) and the extensions discussed below.
 
+The `content/02.delete-me.md` file in the Rootstock repository shows many of the elements and formatting options supported by Manubot.
+See the [raw markdown](https://gitlab.com/manubot/rootstock/blob/master/content/02.delete-me.md#L) in this file and compare it to the [rendered manuscript](https://manubot.github.io/rootstock/).
```
dhimmel: Meant to help address https://github.com/manubot/rootstock/issues/288#issuecomment-567065243.

CC @Nebucatnetzer and @vincerubinetti

dhimmel: Using GitLab link to raw markdown as a workaround for https://github.com/isaacs/github/issues/297. Will be slightly confusing, but not sure what is best.

USAGE.md
agitter: Agreed.  I think that was a remnant from deep review where we were trying to automate things like funding statements.  We haven't needed them in more modern forms of Manubot.

PR #297, opened 2020-01-02
GitHub Actions workflow for building and deployment

dhimmel:
Switch to GitHub actions as the primary CI service for new manuscripts, while retaining backwards compatibility with existing Travis manuscripts.

Todo:

- [x] update manubot python package to detect GitHub CI variables


.github/workflows/manubot.yaml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          name: output
+          path: output
+      - name: Deploy
+        if: (
+          github.ref == 'refs/heads/master' &&
+          github.event_name == 'push' &&
+          job.status == 'success')
+        env:
+          MANUBOT_SSH_PRIVATE_KEY: ${{ secrets.MANUBOT_SSH_PRIVATE_KEY }}
```
dhimmel: perhaps we can also test if `secrets.MANUBOT_SSH_PRIVATE_KEY` is set and if not, provide a better error message

dhimmel:
This action is alive and working (including deployment) at:

- https://github.com/data2health/contributorship
- https://github.com/greenelab/iscb-diversity-manuscript

One question is whether we want to switch to GitHub Actions as the primary method CI method for rootstock and future clones. I think we probably do, since it reduces the setup from using two services (GitHub and Travis) to just one (GitHub). It's also more aligned with the longterm roadmap.

Unfortunately, we still have to manually generate a deploy key to push to GitHub until since `secrets.GITHUB_TOKEN` [only triggers](https://github.com/peaceiris/actions-gh-pages/issues/9) GitHub Pages pages builds on private (not public) repos.

One option if we switch to actions as the primary CI going forward, is that we can add something like the following to the cleanup step of setup:

```shell
git rm .travis.yml .appveyor.yml ci/install.sh
```

Or do we want to remove this files from rootstock, thereby nudging existing users to switch?

.github/workflows/manubot.yaml
```diff
@@ -0,0 +1,40 @@
+name: Manubot
```
vincerubinetti: Maybe this could have a more descriptive name, since the user's might theoretically poke around and see it?

image

Maybe `Manubot Continuous Integration` or `Manubot CI` or just `Manubot Build and Deploy Manuscript`

Maybe the file name could be `manubot-ci.yml` or `manubot-build-and-deploy.yml`

Finally, is "manubot" necessary in these names since we're already in the context of manubot, being in a manubot repository.

vincerubinetti: Maybe you want to add a status badge to the readme:
``

.github/workflows/manubot.yaml
```diff
@@ -0,0 +1,40 @@
...
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  build:
+    name: Build Manuscript
```
vincerubinetti: Maybe `Build and Deploy Manuscript` to be more accurate, and to avoid duplicate name of the `Build Manuscript` step below.

.github/workflows/manubot.yaml
```diff
@@ -0,0 +1,40 @@
...
+        env:
+          GITHUB_PULL_REQUEST_SHA: ${{ github.event.pull_request.head.sha }}
+        shell: bash --login {0}
+        run: bash build/build.sh
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v1
+        with:
+          name: output
+          path: output
+      - name: Deploy
```
vincerubinetti: Maybe `Deploy Manuscript` to keep in line with the `[verb] [noun]` pattern of the other step names above.

ci/deploy.sh
```diff
@@ -7,9 +7,16 @@ set -o errexit \
     -o nounset \
     -o pipefail
 
+# set environment variables for either Travis or GitHub Actions
+REPO_SLUG=${TRAVIS_REPO_SLUG:-$GITHUB_REPOSITORY}
```
vincerubinetti: I assume the `:-` is like an `or`?

I agree with what you mentioned that this isn't super clean. One other option, I think, is to just leave the github ones, and make comments about how to fill them in for other services.

I know Travis is (will be) a legacy thing, but perhaps we should just fill in the variables for the service we recommend by default (github), and instruct the user on what the variables would be for Travis or Circle CI or something else.

I think all of this might become somewhat irrelevant when the major refactor happens anyway, so I'm not too concerned.

.github/workflows/manubot.yaml
vincerubinetti: Oh wait, that wouldn't be for the specific manuscript repo. Is there a way to get that dynamically? Would a relative image link work, like ``

.github/workflows/manubot.yaml
vincerubinetti: FWIW you can also omit this name, and I think it will show up in the `actions` tab as just the name of the whole config file.

vincerubinetti: FWIW I think you can omit this and in the `Actions` tab it will just show up as the config file name at the top.

.github/workflows/manubot.yaml
vincerubinetti: Oh wait, that wont refer to the specific manuscript. Would a relative image link work? ``

dhimmel: > is "manubot" necessary in these names since we're already in the context of manubot

I am thinking about the situation where someone creates a repository with a manuscript as well as other stuff. The other stuff may need additional workflows. Those other workflows may have nothing to do with Manubot... for example perhaps one runs an R analysis, another one spellchecks pull requests, another makes sure PR authors have signed a contributor license agreement, another one automatically merges approved PRs. In these cases, I envision you want workflow names that best differentiate the specific workflow.

Since the name "Manubot" is used in the context of a workflow, I think adding `CI` or "Continuous Integration" afterwards is a bit redundant. My understanding is that workflows are basically CI, and perhaps GitHub wanted to get away from the CI terminology because it's a bit confusing.

One final point is that I see the GitHub action workflow as become the main way people run manubot. The rest of the manuscript repo will be their content / config files. The action is actually "the manubot", or at least the place where "the manubot" runs.

dhimmel: The relative image link won't work, first because it's an SVG and second because that is not a file path relative to the repo content.

We could use the [absolute URL](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository). It would get rebranded to the manubot instance from `manubot/rootstock` during setup. Do you think badges are important given that you can now see the check or fail mark from the main page? I guess it makes it more clear for those unfamiliar with CI / GitHub actions?

ci/deploy.sh
dhimmel: > I assume the `:-` is like an `or`?

If the first variable is unset, return the second. So yes like a Python `or` that returns the actual values rather than True / False.

> One other option, I think, is to just leave the github ones, and make comments about how to fill them in for other services.

I'd like to keep existing Travis CI setups working if we can. The elegance of this `ci/deploy.sh` script is not super important. Users usually do not touch it and hopefully soon, we'll be moving it out of individual manuscript repositories.

.github/workflows/manubot.yaml
agitter: I do think that badges are more obvious to a larger audience that may not know to look for the pass/fail icons.

.github/workflows/manubot.yaml
agitter: Yes, better error messages would be valuable.  Setting up CI has been a common source of errors.

.github/workflows/manubot.yaml
agitter: Yes, I like this suggestion too

.github/workflows/manubot.yaml
```diff
@@ -0,0 +1,40 @@
...
+          activate-environment: manubot
+          environment-file: build/environment.yml
+          auto-activate-base: false
+          miniconda-version: 'latest'
+      - name: Build Manuscript
+        env:
+          GITHUB_PULL_REQUEST_SHA: ${{ github.event.pull_request.head.sha }}
+        shell: bash --login {0}
+        run: bash build/build.sh
+      - name: Upload Artifacts
```
agitter: Will this eventually replace the AppVeyor artifacts and notification messages?

README.md
```diff
@@ -4,7 +4,6 @@
 
 [![HTML Manuscript](https://img.shields.io/badge/manuscript-HTML-blue.svg)](https://manubot.github.io/rootstock/)
 [![PDF Manuscript](https://img.shields.io/badge/manuscript-PDF-blue.svg)](https://manubot.github.io/rootstock/manuscript.pdf)
-[![Build Status](https://travis-ci.com/manubot/rootstock.svg?branch=master)](https://travis-ci.com/manubot/rootstock)
```
agitter: I like the suggestion to include a GitHub Actions badge here.  Because we can't adaptively show the Travis or GitHub Actions badge based on which CI service is being used, would we default to GitHub's?

SETUP.md
```diff
@@ -62,28 +62,36 @@ git push --set-upstream origin output
...
+- [GitHub Actions](https://github.com/features/actions).
+  Configured at [`.github/workflows/manubot.yaml`](.github/workflows/manubot.yaml)
+- [Travis CI](https://travis-ci.com).
+  Configured at [`.travis.yml`](.travis.yml).
+- [AppVeyor](https://www.appveyor.com/).
+  Configured at [`.appveyor.yml`](.appveyor.yml).
+
+GitHub Actions is the default service for new manucripts since it's easiest to setup.
+We recommend using either GitHub Actions or Travis CI, but not both to avoid deploying manuscripts multiple times.
+AppVeyor can be used in addition to GitHub Actions or Travis CI to comment on pull request with download links to renderred PDFs.
```
agitter: "renderred" -> "rendered" here and in the next line.

SETUP.md
```diff
@@ -93,23 +101,61 @@ openssl base64 -A -in deploy.key > deploy.key.txt
...
 Finally, click "Add key".
 
-#### Add the private key to Travis CI
+#### Add the private key to GitHub
+
+```sh
+# Print the URL for adding the private key to GitHub
+echo "https://github.com/$OWNER/$REPO/settings/secrets"
+
+# Print the encoded private key for copy-pasting to Travis CI
```
agitter: "Travis CI" -> "GitHub"

SETUP.md
```diff
@@ -93,23 +101,61 @@ openssl base64 -A -in deploy.key > deploy.key.txt
...
+
+Next, go to the GitHub repository settings page (URL echoed above).
+Click "Add a new secret".
+For "Name", enter `MANUBOT_SSH_PRIVATE_KEY`.
+Next, copy-paste the content of `deploy.key.txt` into "Value"
+(printed above by `cat`, including any trailing `=` characters if present).
+
+### GitHub Actions
+
+If you plan on only using GitHub actions, you can remove configuration files for other CI services:
```
agitter: I think this is fine for now instead of removing the files from rootstock.  Some users may prefer to continue using Travis CI for reasons we can't anticipate.

SETUP.md
```diff
@@ -93,23 +101,61 @@ openssl base64 -A -in deploy.key > deploy.key.txt
...
 
 Go to the GitHub settings URL echoed above in a browser, and click "Add deploy key".
 For "Title", add a description like "Manubot Travis Deploy Key".
-Copy-paste the contents of the `deploy.key.pub` text file (printed above by `cat`) into the "Key" text box.
+Copy-paste the contents of the `ci/deploy.key.pub` text file (printed above by `cat`) into the "Key" text box.
 Check the "Allow write access" box below.
 Finally, click "Add key".
 
-#### Add the private key to Travis CI
+#### Add the private key to GitHub
```
agitter: I didn't test these steps, but they are easy to understand.

.github/workflows/manubot.yaml
dhimmel: The main difference now is that it doesn't provide notification messages (possibly it could not sure). Although I'm not sure I love the messages since they get overwhelming.

So yeah, I think it is a satisfactory substitute, but requires users are aware of the feature, since you have to navigate to it.

README.md
dhimmel: I think it's okay to have both. Travis CI fails gracefully



GitHub's doesn't. Note I'll have to merge to even see whether the GitHub badge works.

.github/workflows/manubot.yaml
dhimmel: Okay I will save this for a subsequent PR. Possibly that also removes the legacy travis deploy key method of deployment.

PR #298, opened 2020-01-04
SETUP: better manubot update instructions

dhimmel:


SETUP.md
```diff
@@ -174,14 +174,18 @@ This process can be difficult, especially if conflicts have arisen, and is recom
...
+# checkout a new branch, named using the current date, i.e. rootstock-2018-11-16
 git checkout -b rootstock-$(date '+%Y-%m-%d')
 ```
 
 Second, pull the new commits from rootstock, but do not automerge:
 
-```sh
+```shell
+# add rootstock remote, if it doesn't exist
+git config remote.rootstock.url >&- || git remote add rootstock https://github.com/manubot/rootstock.git
```
agitter: `>&-` is being used to suppress the output of the `git config` command, right?

dhimmel: I took it from https://stackoverflow.com/a/57936187/4651668

Without `>&-`, if a the rootstock remote is already set, then you get the following output `https://github.com/manubot/rootstock.git`. I'm okay with that output. It's useful to see if rootstock points elsewhere than `manubot/rootstock`. So I will remove the ugly and confusing `>&-`

PR #300, opened 2020-01-07
Process citations / references with pandoc-manubot-cite filter

dhimmel:
Before merge:

- change repo back to `manubot/manubot`

Testing out https://github.com/manubot/manubot/pull/99 and how it would be integrated with rootstock.


content/metadata.yaml
```diff
@@ -25,3 +25,6 @@ authors:
     affiliations:
       - Department of Something, University of Whatever
       - Department of Whatever, University of Something
+pandoc:
+  citekey-aliases:
+    tag:deep-review: doi:10.1098/rsif.2017.0387
```
dhimmel: For reference, it looks like users need to replace the tab separation from `citation-tags.tsv` with a space for the YAML to be valid. We could look into `manubot process` converting citation-tags to citation-aliases if `--skip-citations` is specified.

I'm also not sure that this is the best place to specify `citekey-aliases`. It is okay. We should see how it interacts with defaults and per-markdown file yaml metadata blocks. Another option is to default to the link-reference syntax for defining aliases.

agitter: A possible downside of defaulting to the link-reference syntax could be a greater chance of multiply defined tags.  We could recommend putting them all in one Markdown file, even a new dedicated files like `tags.md`.

dhimmel: > A possible downside of defaulting to the link-reference syntax could be a greater chance of multiply defined tags

This is easy to detect and warn.

> We could recommend putting them all in one Markdown file, even a new dedicated files like tags.md.

 One downside of this is that it doesn't allow normal markdown viewers to render the link. One of the big benefits of link-reference syntax is the ability of:

```markdown
[@citekey] to render as a link

[@citekey]: https://github.com/manubot/rootstock/pull/300
```

To render in places like GitHub like below:

[@citekey] to render as a link

[@citekey]: https://github.com/manubot/rootstock/pull/300

dhimmel:
@agitter I am going to merge, since I want to start rolling out this change for additional testing and usage experience. Will address any comments in a subsequent PRs.

PR #305, opened 2020-01-28
Attempt to version artifacts

dhimmel:
look into addressing https://github.com/czbiohub/orthology-free-comparative-transcriptomics-paper/pull/41#issuecomment-579419055

### DIFF
--- a/.github/workflows/manubot.yaml
+++ b/.github/workflows/manubot.yaml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches:
       - master
+env:
+  GITHUB_PULL_REQUEST_SHA: ${{ github.event.pull_request.head.sha }}
+  TRIGGERING_SHA_8: ${${GITHUB_PULL_REQUEST_SHA:-$GITHUB_SHA}::7}
 jobs:
   manubot:
     name: Manubot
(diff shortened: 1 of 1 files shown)


dhimmel:
@olgabot, this PR will enable the artifacts archive to be named something like `output-121907e`. We could create any name we want like: `2020-01-28-manubot-output-121907e`. @agitter @olgabot, any preferences?

agitter:
I'd say the commit hash is the most essential part of the filename.  The datestamp alone is ambiguous when multiple builds are run in the same day, and a build or job id is not as easy to map back to the source modifications for that build.

Beyond that, I don't have strong opinions.  `2020-01-28-manubot-ci-output-121907e` is a little long but works for me.  `output-121907e` is also good.

I'm glad to see that someone else likes these AppVeyor notifications :)

dhimmel:
What about `manuscript-$GITHUB_ACTION-$COMMIT` like `manuscript-413593601-121907e`?

Things I'd like in the name:

1. sorts chronologically
2. sorts such that all other manubot output archives are adjacent in a cluttered Downloads folder
3. easy to associate output with commit that triggered it and the action that created it
4. recognizable as a manubot manuscript in a Downloads folder where it's been forgotten about.

agitter:
> Things I'd like in the name:

Those are all great properties.  I like the scheme you went with.

PR #310, opened 2020-02-17
GH Actions: no setup deployment using GITHUB_TOKEN

dhimmel:
Evaluates whether we can use `secrets.GITHUB_TOKEN` for deployment. Testing the PR in the repository https://github.com/dhimmel/rootstock-github-token.

Also noting https://github.com/dhimmel/rootstock-actions-deploy to test using [`peaceiris/actions-gh-pages`](https://github.com/peaceiris/actions-gh-pages). However, I think that using our existing `deploy.sh` is less disruptive for now.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -1,7 +1,7 @@
 # Table of contents
 
 - [Table of contents](#table-of-contents)
-- [Cloning the manubot/rootstock repository to create a new manuscript](#cloning-the-manubot-rootstock-repository-to-create-a-new-manuscript)
+- [Creating a new manuscript](#creating-a-new-manuscript)
   * [Configuration](#configuration)
   * [Create repository](#create-repository)
   * [Continuous integration](#continuous-integration)
@@ -17,10 +17,12 @@
 
 _generated with [markdown-toc](https://ecotrust-canada.github.io/markdown-toc/)_
 
-# Cloning the manubot/rootstock repository to create a new manuscript
+# Creating a new manuscript
 
-The process to create a new Manubot manuscript is a bit challenging, because it requires a few steps that are difficult to automate.
+These instructions detail how to create a new manuscript based off of the [`manubot/rootstock`](https://github.com/manubot/rootstock/) repository.
+The process can be a bit challenging, because it requires a few steps that are difficult to automate.
 However, you will only have to perform these steps once for each manuscript.
+
 These steps should be performed in a command-line shell (terminal), starting in the directory where you want the manuscript folder be created.
 Setup is supported on Linux, macOS, and Windows.
 Windows setup requires [Git Bash](https://gitforwindows.org/) or [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/faq).
--- a/SETUP.md
+++ b/SETUP.md
@@ -83,6 +83,12 @@ Manubot integrates with cloud services to perform continuous integration (CI).
 For Manubot that means automatically building and deploying your manuscript.
 Manubot supports the following services:
 
+| Service | Default | Artifacts | Deployment | Config | Private Repos |
+|---------|---------|-----------|---------|--------|---------------|
+| [GitHub Actions](https://github.com/features/actions) | ✔️ | ✔️ | ✔️ | [`manubot.yaml`](.github/workflows/manubot.yaml) | 2,000 minutes per month |
+| [Travis CI](https://travis-ci.com) | ❌ | ❌ | ✔️ needs setup | [`.travis.yml`](.travis.yml) | 100 build trial |
+| [AppVeyor](https://www.appveyor.com/) | ❌ | ✔️ with PR comments | ❌ | [`.appveyor.yml`](.appveyor.yml) | 14 day trial |
+
 - [GitHub Actions](https://github.com/features/actions).
   Configured at [`.github/workflows/manubot.yaml`](.github/workflows/manubot.yaml)
 - [Travis CI](https://travis-ci.com).
--- a/SETUP.md
+++ b/SETUP.md
@@ -95,7 +95,7 @@ Manubot supports the following CI services:
 
 Notes on table fields:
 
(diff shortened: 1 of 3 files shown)


dhimmel:
https://github.com/manubot/rootstock/pull/310/commits/16b1a82da6e01a559184d8821663a3a6ed0594ed properly deployed in [this build](https://github.com/dhimmel/rootstock-github-token/runs/451006688#step:7:26) and triggered the GitHub Pages build despite using GITHUB_TOKEN on a public repo!

SETUP.md
```diff
@@ -95,9 +95,28 @@ We recommend using either GitHub Actions or Travis CI, but not both to avoid dep
...
+```
+
+<details>
+<summary>Expand for Deploy Key</summary>
+
 ### Deploy key
 
-Generate a deploy key so CI can write to the repository.
+Deployment on Travis CI requires a SSH Deploy Key.
+Previousely, GitHub Actions also required an SSH Deploy Key, but now GitHub supports using `secrets.GITHUB_TOKEN`.
```
agitter: `Previousely` -> `Previously`

Do we want to explain the differences between GITHUB_TOKEN, an SSH deploy key, and a GitHub personal access token in terms of permissions?  These terms can be confusing for people who aren't experienced with continuous integration.  Others may mix up this GITHUB_TOKEN with the broader personal access token.

ci/deploy.sh
```diff
@@ -26,8 +26,18 @@ git config --global push.default simple
...
+  echo >&2 "[INFO] Detected SECRET_GITHUB_TOKEN. Will deploy via HTTPS."
+  USE_GITHUB_TOKEN=true
+  git remote set-url origin "https://$SECRET_GITHUB_TOKEN@github.com/$REPO_SLUG.git"
+else
+  echo >&2 "[INFO] Missing SECRET_GITHUB_TOKEN. Will deploy via SSH."
+  USE_GITHUB_TOKEN=false
+  git remote set-url origin "git@github.com:$REPO_SLUG.git"
+fi
+
+if [ $USE_GITHUB_TOKEN = false ]; then
```
agitter: Do you want to indent the next block?

ci/deploy.sh
```diff
@@ -87,6 +98,3 @@ ghp-import \
   --branch=gh-pages \
   --message="$MESSAGE" \
   webpage
-
-# Workaround https://github.com/travis-ci/travis-ci/issues/8082
```
agitter: Was this Travis issue resolved?

dhimmel: We put this here in https://github.com/manubot/rootstock/pull/23. I think it's been resolved upstream with Travis, but we kept it in case there was a Travis regression.

ci/deploy.sh
dhimmel: If we indent the following the two lines this will cause conflicts with legacy users who have modified those variable names:

```shell
  -K $encrypted_9befd6eddffe_key \
  -iv $encrypted_9befd6eddffe_iv \
```

Therefore I propose we fix the indentation when we finally remove the encrypted variable method.

dhimmel:
Note that previously the approach in this PR did not work because GITHUB_TOKEN deployments did not trigger GitHub Pages builds for public repositories. As reported by a community member in https://github.com/peaceiris/actions-gh-pages/issues/9#issuecomment-586552421, this no longer seems to be the case. However, there has been no official GitHub announcement about the change, so there is a risk that it will be temporary.

ci/deploy.sh
agitter: Agreed.  Let's keep it as is.

dhimmel:
https://github.com/manubot/rootstock/pull/310/commits/eddd939c7481bcee6e6c08b08c2d5f8bc2ecd0a7 deployed properly in [this build](https://github.com/dhimmel/rootstock-github-token/runs/451250623#step:7:15).

https://github.com/manubot/rootstock/pull/310/commits/eddd939c7481bcee6e6c08b08c2d5f8bc2ecd0a7 uses `MANUBOT_SSH_PRIVATE_KEY` over `MANUBOT_ACCESS_TOKEN` when both are set. The idea is that if GitHub's support for GITHUB_TOKEN is temporary (since no official commitment yet), then users who have configured `MANUBOT_ACCESS_TOKEN` will not be affected. Only users who did not set up `MANUBOT_SSH_PRIVATE_KEY` will be affected, but they could then set up `MANUBOT_SSH_PRIVATE_KEY` without having to make Manubot Rootstock updates to get deployment to work.

I'll try to improve the setup docs. After that I think this PR will be ready.

SETUP.md
dhimmel: I updated the SETUP extensively, hopefully a bit clearer on these points although still don't go into too much detail.

dhimmel:
@agitter ready for re-review

SETUP.md
```diff
@@ -81,23 +85,57 @@ git push --set-upstream origin output
...
+  Travis CI will only deploy if an SSH Private Key is provided.
+  To avoid deploying a manuscript multiple times, disable GitHub Actions before providing an SSH Private Key to Travis.
+- **Config**: File configuring what operations CI will perform.
+- **Private Repos**: Quota for private repos.
+  Only GitHub Actions supports cost-free builds of private repositories beyond a trial period.
+  All services are cost-free for public repos.
+
+### GitHub Actions
+
+GitHub Actions is the recommended default CI service because it requires no additional setup.
```
agitter: It looks like GitHub Actions is enabled by default, right?  The owner has to explicitly disable it in the repository settings.  If that's accurate, these instructions make sense.

dhimmel: In the most common case, adding the workflow file should be enough to enable actions. However, it is possible that actions could be disabled at the:

1. repo level
2. [organization level](https://github.blog/2019-10-03-github-actions-adds-policy-controls-for-organization-and-repository-owners/)

Not sure how common this will be.

PR #313, opened 2020-02-18
Do not deploy forked repositories

dhimmel:
It looks like GitHub actions was also running on my fork. Now GitHub Actions deploys even without a manually set SSH Key. Therefore, it [tried to deploy on my fork](https://github.com/dhimmel/manubot-rootstock/commit/d14dd7dc8503fdae86c1a5f66777b62e11a0ebc6/checks?check_suite_id=465620679#step:7:17).

### DIFF
--- a/.github/workflows/manubot.yaml
+++ b/.github/workflows/manubot.yaml
@@ -9,7 +9,6 @@ on:
 jobs:
   manubot:
     name: Manubot
-    if: '!github.repository.fork'
     runs-on: ubuntu-latest
     env:
       GITHUB_PULL_REQUEST_SHA: ${{ github.event.pull_request.head.sha }}
--- a/.github/workflows/manubot.yaml
+++ b/.github/workflows/manubot.yaml
@@ -9,6 +9,7 @@ on:
 jobs:
   manubot:
     name: Manubot
+    if: '!github.repository.fork'
     runs-on: ubuntu-latest
     env:
       GITHUB_PULL_REQUEST_SHA: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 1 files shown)


dhimmel:
When will users want to deploy a manuscript to GitHub Pages? In general, I think they'll only want to deploy the upstream manuscript in case of forks. Although this may not always be the case. Sometimes a fork may want to deploy as well. Perhaps we could default to only deploying non-fork repos with a way to override and deploy a fork.

agitter:
I agree that the default should be that forks do not deploy.  Users may want to build the manuscript in a fork to test changes as long as it doesn't deploy.

dhimmel:
> Users may want to build the manuscript in a fork to test changes as long as it doesn't deploy.

The problem is with our current configuration, we have:

```yaml
on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - master
```

So we're not building branches on forks, which is the main place users would be testing these changes.

I propose merging this PR as a quick fix since it will prevent failing fork master builds. And then thinking more about the ideal configuration.

dhimmel:
For reference, I dumped the github actions context in this workflow run https://github.com/dhimmel/rootstock-actions-deploy/runs/453060191

dhimmel:
Another possibility is that we could use the `github.repository.has_pages` context variable like:

```
'!github.repository.fork || github.repository.has_pages
```

> Works for me. I wasn't sure why both conditional statements are needed, but if it has the intended behavior that's good enough for me.

I will remove the first. Was for testing, but didn't actually test anything.

agitter:
What's the default state of `github.repository.has_pages` for a forked repository?  Is that false initially?

dhimmel:
See https://github.com/dhimmel/manubot-rootstock/actions/runs/41309705 for the fork action on the merged commit.

agitter:
It still tried to deploy?

dhimmel:
> What's the default state of `github.repository.has_pages` for a forked repository? Is that false initially?

It seems that forking on github does replicate the `gh-pages` and `output` branches. However, the GitHub Pages build does not run the branch until another commit is made such that there is not webpage available. Not sure what has_pages would be but guessing it's true. 


> It still tried to deploy?

ahh. will debug

dhimmel:
Hopefully fixed in https://github.com/dhimmel/manubot-rootstock/commit/cd8ce12f739c4ea5981723bade4f1d8c8cadaf1a. If the deployment is properly skipped I will push this commit to `manubot/rootstock:master`

PR #315, opened 2020-02-27
Update environment and usage on 2020-02-27

dhimmel:
todo: update usage

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -76,9 +76,15 @@ We recommend always specifying the width of SVG images (even if just `width="100
 
 ### Citations
 
-Manubot supports Pandoc [citations](https://pandoc.org/MANUAL.html#citations) via `pandoc-citeproc`.
-However, Manubot performs automated citation processing and metadata retrieval on in-text citations.
-Therefore, citations must be of the following form: `@source:identifier`, where `source` is one of the options described below.
+Manubot supports Pandoc [citations](https://pandoc.org/MANUAL.html#citations).
+Citations are processed in 3 stages:
+
+1. Pandoc parses the input Markdown to locate citation keys.
+2. The [`pandoc-manubot-cite`](https://github.com/manubot/manubot#pandoc-filter) filter automatically retreives the bibliographic metadata for citation keys.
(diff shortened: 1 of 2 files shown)


USAGE.md
```diff
@@ -109,13 +109,21 @@ Note that multiple citations must be semicolon separated.
 Be careful not to cite the same study using identifiers from multiple sources.
 For example, the following citations all refer to the same study, but will be treated as separate references: `[@doi:10.7717/peerj.705; @pmcid:PMC4304851; @pmid:25648772]`.
 
+Citation keys must adhere to the syntax described in the [Pandoc manual](https://pandoc.org/MANUAL.html#citations):
```
agitter: The text here explains the current behavior well, and the online regex is an especially helpful resource.  Because this is a breaking change, should we more prominently note that as of `<date>` the Manubot citation behavior changed?  I don't know how many authors besides us regularly update rootstock.  I imagine they could be confused, however, if they update and a previously working manuscript starts to fail due to the citation key syntax changes.

USAGE.md
```diff
@@ -109,13 +109,21 @@ Note that multiple citations must be semicolon separated.
...
+
+> The citation key must begin with a letter, digit, or `_`, and may contain alphanumerics, `_`, and internal punctuation characters (`:.#$%&-+?<>~/`). Here are some examples:
```
agitter: We can remove `Here are some examples:` because the examples aren't copied here.

USAGE.md
```diff
@@ -109,13 +109,21 @@ Note that multiple citations must be semicolon separated.
...
+
+To evaluate whether a citation key fully matches this syntax, try [this online regex](https://regex101.com/r/mXZyY2/latest).
+If the citation key is not valid, use the [citation tag](#citation-tag) workaround below.
+This is required for citation keys that contain forbidden characters such as `;` or `=` or end with a non-alphanumeric character such as `/`.
+See [jgm/pandoc#6026](https://github.com/jgm/pandoc/issues/6026) for progress on a more flexible Markdown citation key syntax.
```
agitter: I'm not opposed to tracking this progress here, but it seems more relevant for developers than Manubot users.

PR #316, opened 2020-02-29
GitHub Actions: enable caching

dhimmel:
https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#using-the-cache-action

### DIFF
--- a/.github/workflows/manubot.yaml
+++ b/.github/workflows/manubot.yaml
@@ -24,6 +24,13 @@ jobs:
         with:
           # fetch entire commit history to support get_rootstock_commit
           fetch-depth: 0
+      - name: Cache
+        uses: actions/cache@v1
+        with:
+          path: ci/cache
+          key: ci-cache-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/manubot.yaml
```diff
@@ -24,6 +24,13 @@ jobs:
         with:
           # fetch entire commit history to support get_rootstock_commit
           fetch-depth: 0
+      - name: Cache
+        uses: actions/cache@v1
+        with:
+          path: ci/cache
+          key: ci-cache-${{ github.ref }}
+          restore-keys: |
+            ci-cache-refs/heads/master
```
dhimmel: the point of these three lines is to make it so that PRs can never write to the master branch cache but can still read from it if they don't have their own cache yet

dhimmel:
Note that we might also be able to cache the conda installation as per https://github.com/goanpeca/setup-miniconda/issues/13.

PR #333, opened 2020-04-20
Enable pandoc spellcheck filter

agitter:
References https://github.com/manubot/manubot/issues/31
Initial implementation from https://github.com/agitter/my-manuscript/pull/15

There are a few open questions to discuss:
- [x] Where should this be documented (`SETUP.md`?)
- [x] Should it be enabled in AppVeyor builds by default (I vote yes)
- [x] Should it be enabled in Travis CI builds by default?
- [x] What should be in the initial custom dictionary?

I've noticed that many spelling errors come from `00.front-matter.md`.  We could automatically add all of the manuscript metadata to the custom dictionary, but then we would ignore typos in affiliations or titles.  I would prefer to save that for a separate pull request even if we do want it.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -85,7 +85,10 @@ fi
 # Spellcheck
 if [ "${SPELLCHECK:-}" = "true" ]; then
   export ASPELL_CONF="add-extra-dicts $(pwd)/build/assets/custom-dictionary.txt; ignore-case true"
-  pandoc --lua-filter spellcheck.lua output/manuscript.md | uniq | while read word; do grep -on "\<$word\>" content/*md; done | sort -h -t ":" -k 1b,1 -k2,2 > output/spelling-errors.txt
+  # Use "|| true" after grep because otherwise this step of the pipeline will
+  # return exit code 1 if any of the markdown files do not contain a
+  # misspelled word
+  pandoc --lua-filter spellcheck.lua output/manuscript.md | uniq | while read word; do grep -ion "\<$word\>" content/*.md; done || true | sort -h -t ":" -k 1b,1 -k2,2 > output/spelling-errors.txt
   cat output/spelling-errors.txt
 fi
 
--- a/build/build.sh
+++ b/build/build.sh
@@ -84,7 +84,7 @@ fi
 
 # Spellcheck
 if [ "${SPELLCHECK:-}" = "true" ]; then
-  export ASPELL_CONF="add-extra-dicts $(pwd)/build/assets/custom-dictionary.txt"
+  export ASPELL_CONF="add-extra-dicts $(pwd)/build/assets/custom-dictionary.txt; ignore-case true"
   pandoc --lua-filter spellcheck.lua output/manuscript.md | sort | uniq > output/spelling-errors.txt
   cat output/spelling-errors.txt
 fi
--- a/build/build.sh
+++ b/build/build.sh
@@ -85,7 +85,7 @@ fi
 # Spellcheck
 if [ "${SPELLCHECK:-}" = "true" ]; then
   export ASPELL_CONF="add-extra-dicts $(pwd)/build/assets/custom-dictionary.txt; ignore-case true"
-  pandoc --lua-filter spellcheck.lua output/manuscript.md | sort | uniq > output/spelling-errors.txt
+  pandoc --lua-filter spellcheck.lua output/manuscript.md | uniq | while read word; do grep -on "\<$word\>" output/manuscript.md; done | sort -h > output/spelling-errors.txt
   cat output/spelling-errors.txt
 fi
 
--- a/build/build.sh
+++ b/build/build.sh
@@ -85,7 +85,7 @@ fi
 # Spellcheck
 if [ "${SPELLCHECK:-}" = "true" ]; then
   export ASPELL_CONF="add-extra-dicts $(pwd)/build/assets/custom-dictionary.txt; ignore-case true"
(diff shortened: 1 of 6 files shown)


agitter:
The first draft of the spellcheck pull request is working in AppVeyor and Travis CI.  Travis CI doesn't store the spelling errors as an artifact or add them as a pull request comment.  However, it can dump them in the build log.

dhimmel:
Really great progress! Will help catch spelling errors hopefully.

Did you look into detecting the language set in `metadata.lang` and using it for aspell?

It'd be nice to show spellcheck output as a github actions step. Then it'd be easy to expand that section of the output to check for spelling errors. Was the motivation for excluding Actions that it doesn't use `ci/install.sh`?

> Should it be enabled in AppVeyor builds by default (I vote yes)

I think so. The only concern I'd have is non-english manuscripts, but it's not too hard to turn off or switch the language.

>  Should it be enabled in Travis CI builds by default?

Doesn't matter. I view Travis CI support mostly for backwards compatibility at this point and less about new features.

> What should be in the initial custom dictionary?

From https://github.com/agitter/my-manuscript/pull/15#issuecomment-615864322, it looks like lot's of manubot related terms cause problems: LaTeX, ORCID, permalink... not sure how many are because this is on delete-me which uses words like DOI outside of citekeys.

agitter:
> Did you look into detecting the language set in metadata.lang and using it for aspell?

I haven't, but I will.  I believe this would involve downloading the appropriate aspell dictionary with apt-get and then specifying the language in the `ASPELL_CONF` variable.

> Was the motivation for excluding Actions that it doesn't use ci/install.sh?

I initially planned to only set it up for AppVeyor because it can preview spelling errors in the pull request comment.  If we enable it for GitHub Actions, would you recommend a new [conditional](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions) step between `Install Environment` and `Build Manuscript`?

I should be able to come up with a reasonable custom dictionary based on the errors from delete-me.

agitter:
It might be better to save language detection for a follow up pull request.  Aspell [dictionary files](https://ftp.gnu.org/gnu/aspell/dict/0index.html) use ISO 639 language codes.  An exception is Portuguese, which has `pt_BR` and `pt_PT`, which are not standard.  Pandoc follows the BCP 47 standard for languages, and I haven't yet assessed their [differences](https://en.wikipedia.org/wiki/IETF_language_tag#Relation_to_other_standards).

The first step would be reading the language from `metadata.lang`.  Then, we could install the requested language dictionary and set that as the main dictionary with Asepll's `master` setting.  However, there would be several special cases, string manipulation, and error handling required because.

dhimmel:
> If we enable it for GitHub Actions, would you recommend a new conditional step between Install Environment and Build Manuscript?

I was thinking a conditional step that writes the spellcheck errors file to standard output (if the file exists. But whatever you think is best (including nothing) is okay.

build/build.sh
```diff
@@ -82,4 +82,11 @@ if [ "${BUILD_DOCX:-}" = "true" ]; then
     --defaults="$PANDOC_DEFAULTS_DIR/docx.yaml"
 fi
 
+# Spellcheck
+if [ "${SPELLCHECK:-}" = "true" ]; then
+  export ASPELL_CONF="add-extra-dicts $(pwd)/build/assets/custom-dictionary.txt"
+  pandoc --lua-filter spellcheck.lua output/manuscript.md | sort | uniq > output/spelling-errors.txt
```
danich1, line 88: Can we change 
```bash 
pandoc --lua-filter spellcheck.lua output/manuscript.md | sort | uniq > output/spelling-errors.txt
``` 
to the following line:
```bash
pandoc --lua-filter spellcheck.lua output/manuscript.md | uniq | while read word; do grep -on "\<$word\>" output/manuscript.md; done | sort -h > output/spelling-errors.txt
```
^ This adds line numbers to each misspelled word. Would be useful to identify the exact spot of the error.

agitter: Good idea @danich1.  I tested that in 14215d36694d960475f30c6eab334237a8e43ec0 and the output is below.

Is this line numbering useful enough? The line numbers correspond to `output.md`, but authors will have to edit the original `*.md` files in the `content` subdirectory to correct the spelling errors.

danich1: Good point. This fix appends file names at the beginning of each "misspelled" word:
```bash
pandoc --lua-filter spellcheck.lua output/manuscript.md | uniq | while read word; do grep -on "\<$word\>" content/*md; done | sort -h -t ":" -k 1,2 > output/spelling-errors.txt
```
^ Assuming the desire is to only spell check md files.

agitter: Thanks @danich1.  I added that in a63502f but the CI build failed.  I like this solution, but I don't think I'll be able to debug locally until tomorrow.

agitter:
> Up to you if you want to pursue anything else in this PR or leave it for later.

Thanks for re-reviewing @dhimmel.  My plan is to get @danich1's line numbering idea to work and then merge this.  I added some light documentation to `USAGE.md`.

I'll close https://github.com/manubot/manubot/issues/31 after merging this and create a new rootstock issue with the remaining spellcheck improvements, including GitHub Actions.

agitter:
The build passes now, but I get different output from the Travis CI and AppVeyor builds for a63cb12.

agitter:
Using this with a complex manuscript revealed some bugs: https://github.com/greenelab/covid19-review/pull/261

I'll continue testing there.  In the meantime, I don't recommend this for general use in other manuscripts.

Issue #336, opened 2020-04-29
Spellcheck improvements


agitter:
#333 introduced the first version of a spellchecker, but there are several remaining areas for improvement.

**GitHub Actions:**
We should be able to run the spellcheck filter in GitHub Actions as well and print the misspelled words and their locations in the build log.  In my initial attempt, the `spellcheck.lua` file downloaded in an earlier step in the workflow was not available in the "Build Manuscript" step.  I'll need to debug that.

**Customizing the language:**
We could use the language specified in `metadata.lang` to determine which Aspell dictionary to download and use as the master dictionary.  See https://github.com/manubot/rootstock/pull/333#issuecomment-619259591.

**Discrepancies between spellcheck filter and grep:**
Once this is used with real manuscripts, we should watch for different behaviors between the spellcheck filter and grep call used to locate misspelled words.  The filter is run on `output.md` but grep runs on the input `*.md` files.  Template variables and other details are handled differently.

agitter:
I read through the lua filter code, and it does detect the languages of the Pandoc document and pass them to aspell with the `-l` argument.  Therefore, to support multiple languages, we should only need to install the correct dictionaries.

cgreene:
Make case sensitive an option. Currently requires editing two lines:
https://github.com/manubot/rootstock/blob/master/build/build.sh#L90
and
https://github.com/manubot/rootstock/blob/master/build/build.sh#L112

In the first case setting `ignore-case` to false. In the second removing the `-i` flag from the grep command.

PR #337, opened 2020-04-29
Fix spellcheck bug and track unique spelling errors

agitter:
#333 calculated unique words incorrectly.  `uniq` expects a sorted file.  This update also stores the list of unique misspelled words as a separate artifact, which is helpful for updating the custom dictionary.

We may want to use the spellchecker live in https://github.com/greenelab/covid19-review for a while and keep this open in case I discover other unexpected behavior.


agitter:
https://github.com/greenelab/covid19-review/pull/261 revealed some problems with the spellcheck filter.  The corrections are being applied here.

One issue is caused by the lua filter (https://github.com/pandoc/lua-filters/issues/90).  Only the suffixes of misspelled words are returned if the words contain apostrophes, accents, or other non-letter characters.  I will make a pull request if the filter maintainers want it.  In the meantime, I created a fork that has the behavior we need here.

agitter:
The latest commits enable spellchecking in GitHub Actions and fix a punctuation discrepancy between the spellcheck filter on the output manuscript markdown file and grep on the original markdown.  Now the locations of misspelled words with apostrophes can be reported.

Spellcheck is currently enabled by default in GitHub Actions for testing. We could set the environment variable to false if we want it to be opt in.

I don't plan to work on anything else from #336 immediately.  This could be a good time to review, or I can continue testing spellcheck in the COVID-19 manuscript.

ci/install-spellcheck.sh
```diff
@@ -0,0 +1,11 @@
...
+
+## install-spellcheck.sh: run during a CI build to install Pandoc spellcheck dependencies.
+
+# Set options for extra caution & debugging
+set -o errexit \
+    -o pipefail
+
+sudo apt-get update -y
+sudo apt-get install -y aspell aspell-en
+wget https://raw.githubusercontent.com/pandoc/lua-filters/13c3fa7e97206413609a48a82575cb43137e037f/spellcheck/spellcheck.lua
```
dhimmel, line 11: I wonder if we want to download `spellcheck.lua` to `build/assets` or `build/pandoc-filters`? Perhaps we should also gitignore whatever file this ends up at in case someone runs this locally.

ci/install-spellcheck.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
dhimmel, line 1: have we set the executable bit on this file?

agitter: Good catch. Fixed in 969810341fd4c0095899bae6b2d05812098a0ee6.

ci/install-spellcheck.sh
agitter: This is also a good idea.  Do you know where pandoc searches for filter files?  I haven't been able to find that documented, and it fails to find `build/assets/spellcheck.lua`.

dhimmel: from https://pandoc.org/MANUAL.html

In order of preference, pandoc will look for Lua filters in

1.  a specified full or relative path (executable or non-executable)
2.  `$DATADIR/filters` (executable or non-executable) where `$DATADIR` is the user data directory (see [`--data-dir`](https://pandoc.org/MANUAL.html#option--data-dir), above).

Makes me think we should create a `build/assets/pandoc` directory with a `defaults` and `filter` subdirectory. I can look into this in a new PR. For now you could specify the complete relative path like `build/assets/spellcheck.lua`?

dhimmel: > I can look into this in a new PR. 

done in https://github.com/manubot/rootstock/pull/340, but I think these PRs can be merged separately

agitter: Thanks for starting the pandoc file refactoring.  I'll review it later.

Now that I added the complete relative path, the tests pass again.  This should be ready to merge if you agree.

Issue #334, opened 2020-04-22
Themes


thorowitz:
The Manubot manuscript promises that

> The build process can also accommodate different “themes”, which change the general aesthetics and appearance of the exported document (e.g. from a contemporary sans-serif style to a more traditional serif style). 

Are there existing alternate themes available? Where would I find them?

dhimmel:
Currently, we only have a single default HTML theme [available](https://github.com/manubot/rootstock/tree/073ce0e4e81c82e1ce8c55caeab0ebc6fd0f840c/build/themes). Multiple themes has been a common user request (see https://github.com/manubot/rootstock/issues/112), so we'd like to develop alternative themes in the future.

@thorowitz if there are specific tweaks you'd like to the theme, then let us know here. Many things are possible with small changes to [`build/themes/default.html`](https://github.com/manubot/rootstock/blob/073ce0e4e81c82e1ce8c55caeab0ebc6fd0f840c/build/themes/default.html). Happy to help you figure out how to tweak the theme to your needs.

vincerubinetti:
Luckily changing the font is very easy. You can import online fonts, like [Google Fonts](https://fonts.google.com/), right at the top of `default.html` [like this](https://github.com/manubot/rootstock/blob/master/build/themes/default.html#L5-L6). Then to apply the font to everything, replace "Open Sans" on [this line](https://github.com/manubot/rootstock/blob/master/build/themes/default.html#L15) with your desired font name. You can put multiple fonts there, separated by comma, and for each character of text in your document, the browser will try the left-most font first, then try the next if the it isn't installed on the user's system or if it font doesn't have that particular text character, and so on.

thorowitz:
I was able to change the font, as described. However, I have not been able to change the CSL style. I think the directions in comment #242 may be out of date. The `build.sh` file does not have the line `CSL_PATH=build/assets/style.csl `. Instead, it just loads the Pandoc defaults: `PANDOC_DEFAULTS_DIR="${PANDOC_DEFAULTS_DIR:-build/pandoc-defaults}"` (line 25). I tried altering the `pandoc-defaults`, but that failed for some unknown reason. Then I tried adding the CSL line to build.sh, right after the Pandoc default line: `CSL_PATH=https://www.zotero.org/styles/apa.csl`. But this had no effect on the output.

agitter:
[...]

PR #338, opened 2020-04-30
USAGE: howto set citation style using pandoc defaults

dhimmel:
closes https://github.com/manubot/rootstock/issues/334


thorowitz:
Looks good to me. My only caveat is that the first time I tried using a URL it did not work, so I will have to try that option again.

dhimmel:
> My only caveat is that the first time I tried using a URL it did not work, so I will have to try that option again.

Good to know. If you have any more problems with the URL method, let us know the URL and we'll look into it!

PR #340, opened 2020-05-05
 move build/pandoc-defaults to build/pandoc/defaults

dhimmel:


dhimmel:
we can do https://github.com/manubot/rootstock/pull/340/commits/c3880d6e683c0b9cade3a7e7eac0713c16782821 with or without the additional https://github.com/manubot/rootstock/pull/340/commits/c813d70525041a1401b9264506965e25e9724ee7.

One effect of setting `--data-dir` is that is will prevent Manubot from relying of user-specific local files that are not part of the repository. It also might be more difficult to replace just defaults files and not other files like filters. But overall the pandoc commands are a bit simplified.

build/build.sh
```diff
@@ -31,8 +31,9 @@ mkdir -p output
 # https://pandoc.org/MANUAL.html
 echo >&2 "Exporting HTML manuscript"
 pandoc --verbose \
-  --defaults="$PANDOC_DEFAULTS_DIR/common.yaml" \
-  --defaults="$PANDOC_DEFAULTS_DIR/html.yaml"
+  --data-dir="$PANDOC_DATA_DIR" \
+  --defaults=common \
```
agitter, line 35: Even though the `.yaml` extension is optional, could leaving it specified make it more obvious to users how they can edit these defaults files?

dhimmel: done in https://github.com/manubot/rootstock/pull/340/commits/9b06d3ea990fa1fcb4fb3d19d70b049ec113f6fb

PR #342, opened 2020-05-20
Update manubot for more flexible citation processing

dhimmel:
todo:

- [x] docs
- [x] switch dhimmel/manubot to manubot/manubot once https://github.com/manubot/manubot/pull/225 is merged
- [x] closes https://github.com/manubot/manubot/issues/120: switches from tag to alias terminology

## draft commit message

merges https://github.com/manubot/rootstock/pull/342

Includes major updates to how citations are processed by the pandoc-manubot-cite filter.
See the following commit messages for more information:

- https://github.com/manubot/manubot/commit/7055bcc6524fdf1ef97d838cf0158973e2061595
- https://github.com/manubot/manubot/commit/47b03e0c202e6a0561e6548ba08a4ad790f56eee

Removes requirement to prefix certain citekeys with raw: or tag:
Removes support for deprecated `content/citation-tags.tsv`.
Switches from tag to alias terminology for citation aliases. 
closes https://github.com/manubot/manubot/issues/120

Removes pandas and jsonref dependencies, which are no longer needed.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -106,7 +106,7 @@ When choosing which source to use for a citation, we recommend the following ord
    Manubot uses the Identifiers.org Resolution Service to support [hundreds of prefixes](https://github.com/manubot/manubot/blob/7055bcc6524fdf1ef97d838cf0158973e2061595/manubot/cite/handlers.py#L122-L831).
    For example, citing `@clinicaltrials:NCT04280705` will produce the same bibliographic metadata as `@https://identifiers.org/clinicaltrials:NCT04280705` or `@https://clinicaltrials.gov/ct2/show/NCT04280705`.
 9. For references that do not have any of the above persistent identifiers, the citation key does not need to include a prefix.
-   Citing `@old-manuscript` will work, but only if reference metadata is provided manually.
+   Citing `@old-manuscript` will work, but only if reference metadata is [provided manually](#reference-metadata).
 
 Cite multiple items at once like:
 
--- a/USAGE.md
+++ b/USAGE.md
@@ -80,11 +80,11 @@ Manubot supports Pandoc [citations](https://pandoc.org/MANUAL.html#citations).
 Citations are processed in 3 stages:
 
 1. Pandoc parses the input Markdown to locate citation keys.
-2. The [`pandoc-manubot-cite`](https://github.com/manubot/manubot#pandoc-filter) filter automatically retreives the bibliographic metadata for citation keys.
+2. The [`pandoc-manubot-cite`](https://github.com/manubot/manubot#pandoc-filter) filter automatically retrieves the bibliographic metadata for citation keys.
 3. The [`pandoc-citeproc`](https://github.com/jgm/pandoc-citeproc/blob/master/man/pandoc-citeproc.1.md) filter renders in-text citations and generates styled references.
 
-When using Manubot, citation keys should be formatted like `@source:identifier`,
-where `source` is one of the options described below.
+When using Manubot, citation keys should be formatted like `@prefix:accession`,
+where `prefix` is one of the options described below.
 When choosing which source to use for a citation, we recommend the following order:
 
 1. DOI (Digital Object Identifier), cite like `@doi:10.15363/thinklab.4`.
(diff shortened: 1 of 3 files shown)


dhimmel:
pdf artifact from https://github.com/manubot/rootstock/runs/692971191 / https://github.com/manubot/rootstock/pull/342/commits/93cb0ad0fab220a05113833eff3580853406071b looks good:

[manuscript.pdf](https://github.com/manubot/rootstock/files/4657223/manuscript.pdf)

agitter:
I'll review this soon.  In the meantime, do we also want to update `content/02.delete-me.md` to demonstrate current recommended citation practices for pmid, url, tag, etc.?  And the `id` in `content/manual-references.json`?

dhimmel:
> And the id in content/manual-references.json?

https://github.com/manubot/rootstock/blob/8b9b5ced2c7c963bf3ea5afb8f31f9a4a54ab697/content/manual-references.json#L3

Turns out we're not actually citing this in delete-me. `url:https://github.com/manubot/rootstock` is actually the standard citekey here. I've been switching docs to bare `@https` citations because I think it's cleaner, but `url:` prefix gets added during standardization. For some reason, it was easier to implement it this way.

I like to avoid changing `content/manual-references.json` because of conflicts. One option would be to delete it entirely.

USAGE.md
```diff
@@ -76,44 +76,47 @@ We recommend always specifying the width of SVG images (even if just `width="100
...
+   Similarly, `@https://doi.org/10.1101/142760` is a [workaround](https://github.com/manubot/manubot/issues/16) to set the journal name of bioRxiv preprints to _bioRxiv_.
 7. Wikidata Items, cite like `@wikidata:Q50051684`.
-   Note that anyone can edit or add records on [Wikidata](https://www.wikidata.org), so users are encouraged to contribute metadata for hard-to-cite works to Wikidata as an alternative to using a `raw` citation.
-8. For references that do not have any of the persistent identifiers above, use a raw citation like `@raw:old-manuscript`.
-   Metadata for raw citations must be provided manually.
+   Note that anyone can edit or add records on [Wikidata](https://www.wikidata.org), so users are encouraged to contribute metadata for hard-to-cite works to Wikidata.
+8. Any other compact identifier supported by <https://identifiers.org>.
+   Manubot uses the Identifiers.org Resolution Service to support [hundreds of prefixes](https://github.com/manubot/manubot/blob/7055bcc6524fdf1ef97d838cf0158973e2061595/manubot/cite/handlers.py#L122-L831).
```
agitter, line 106: Is https://registry.identifiers.org/registry a more user-friendly link to use here?  It has descriptions of the identifers.

USAGE.md
```diff
@@ -76,44 +76,47 @@ We recommend always specifying the width of SVG images (even if just `width="100
...
-Here is a sentence with several citations [@doi:10.15363/thinklab.4; @pmid:26158728; @arxiv:1508.06576; @isbn:9780394603988].
+Here is a sentence with several citations [@doi:10.15363/thinklab.4; @pubmed:26158728; @arxiv:1508.06576; @isbn:9780394603988].
 ```
 
 Note that multiple citations must be semicolon separated.
 Be careful not to cite the same study using identifiers from multiple sources.
-For example, the following citations all refer to the same study, but will be treated as separate references: `[@doi:10.7717/peerj.705; @pmcid:PMC4304851; @pmid:25648772]`.
+For example, the following citations all refer to the same study, but will be treated as separate references: `[@doi:10.7717/peerj.705; @pmc:PMC4304851; @pubmed:25648772]`.
 
 Citation keys must adhere to the syntax described in the [Pandoc manual](https://pandoc.org/MANUAL.html#citations):
```
agitter, line 122: In the line below
> If the citation key is not valid, use the [citation tag](#citation-tag) workaround below.

"tag" and the anchor should be updated.

USAGE.md
```diff
@@ -134,46 +137,42 @@ pandoc:
   manubot-fail-on-errors: True
 ```
 
-#### Citation tags
+#### Citation aliases
 
-The system also supports citation tags, which map from one citation key (an alias) to another.
+The system also supports citation aliases, which map from one citation key (the "alias" or "tag") to another.
 Tags are recommended for the following applications:
```
agitter, line 143: "Aliases are recommended"

USAGE.md
```diff
@@ -134,46 +137,42 @@ pandoc:
...
-If `citation-tags.tsv` defines the tag `study-x`, then this study can be cited like `@tag:study-x`.
-This method is deprecated.
-
 ## Reference metadata
 
 Manubot stores the bibliographic details for references (the set of all cited works) as CSL JSON ([Citation Style Language Items](http://citeproc-js.readthedocs.io/en/latest/csl-json/markup.html#csl-json-items)).
-For all citation sources besides `raw`, Manubot automatically generates CSL JSON.
+Manubot automatically generates CSL JSON for most persistent identifiers (as described in [Citations](#citations) above).
 In some cases, automatic metadata retrieval fails or provides incorrect or incomplete information.
 Errors are most common for `url` references.
```
agitter, line 177: I also anticipate many errors for identifiers.org references, but we don't necessarily need to state that here.

USAGE.md
dhimmel: I added both. There are some corner cases where manubot may not support what shows up there, so I want to point users to the actual list manubot uses for debugging purposes.

dhimmel:
> Edit: are they related to manubot/manubot#234?

yes, I think I need to keep the jsonref dereferencing of the CSL JSON schema. Not sure why our manubot tests don't trigger the issue. Will investigate. Luckily this doesn't seem to effect the master branch of rootstock.

dhimmel:
The issue is the requests cache. An old version of https://github.com/dhimmel/csl-schema/raw/manubot/csl-data.json must have been used which required jsonref. When I cleared the travis cache and reran, things worked. I am trying to figure out whether it is only this PR that will have the issue or whether all manuscripts might have it upon upgrading.

Issue #293, opened 2019-12-22
Default CSS headers


Nebucatnetzer:
I noticed that the headings sizes aren't correlating to the content.
E.g. an h5 heading is smaller than the caption of a table.
Since we should start with h2 the headers should be a bit bigger to not get so small when we use h5 as a level 4 heading.

dhimmel:
Here is what github does:

# Heading 1

## Heading 2

### Heading 3

#### Heading 4

##### Heading 5

###### Heading 6

So it seems that GitHub does get smaller and smaller headings, eventually smaller than regular text.

> Since we should start with h2 the headers should be a bit bigger to not get so small when we use h5 as a level 4 heading.

I'm not sure I've ever had 4 levels of in-manuscript levels. When we're back from holidays, @vincerubinetti can perhaps suggest where to change the CSS to increase the level 5 heading size and see if he has any style suggestions.

Nebucatnetzer:
The 4 levels happen because we had to use the project management phases in the document, otherwise 3 levels would be enough.
Smaller than the body text might be okay, what I find strange is that they are smaller than the captions which are usually footnote size?

vincerubinetti:
Relevant:
https://stackoverflow.com/questions/55696808/why-do-h5-and-h6-have-smaller-font-sizes-than-p-in-most-user-agent-default

It is built into HTML/CSS itself, and is not related to Github specifically. It is a long-standing default styling, which I happen to think was a stupid choice. I'd be happy to break with precedence in this case.

Issue #281, opened 2019-11-21
Hypothesis improvements


vincerubinetti:
It seems like we need a bunch of Hypothesis tweaks, and it'd be good to do them all at once. Here's some changes we can think about making:

- have a button to completely unload and hide Hypothesis until the user explicitly opts-in
- add button to disable highlights when printing (it's conceivable that someone would want to print with their Hypothesis highlights, so I think it should be an option)
- modify sidebar appearance to match with manubot default theme better

Combines and addresses: #111 #280

vincerubinetti:
I've created an issue on the Hypothesis client here:
https://groups.google.com/a/list.hypothes.is/forum/#!searchin/dev/disable%7Csort:date/dev/x6Fbyey5MFw/1rJeFQDUAAAJ

Until they add this feature officially, it may be difficult/problematic to have a true on/off toggle button.

dhimmel:
> Until they add this feature officially, it may be difficult/problematic to have a true on/off toggle button.

Would be great to full on/off toggle. If that is not an option, toggle-on alone would still improve the user experience. The idea would be as follows:

1. hypothesis off by default. this would make it harder for #306 to occur. It also would avoid the slow page loads on localhost due to spurious annotations. Ideally the UI element to activate hypothesis could show how many annotations existed... so users can avoid activating when there's nothing to see. I'll look into whether this is possible.

2. user activates hypothesis. Sidebar appears. If it's able to be deactivated... great. If not, a refresh would still remove it.

dhimmel:
> Ideally the UI element to activate hypothesis could show how many annotations existed... so users can avoid activating when there's nothing to see. I'll look into whether this is possible.

Using the [api v1 search enpoint](https://h.readthedocs.io/en/latest/api-reference/v1/#tag/annotations/paths/~1search/get):

https://api.hypothes.is/api/search?limit=0&url=https://lelaboratoire.github.io/rethink-prs-ms/

```json
{"total": 122, "rows": []}
```

vincerubinetti:
Hypothesis seems to have recently refreshed the look of their sidebar, and I think it looks better now, so I don't feel the need to rewrite their CSS right now.

vincerubinetti:
> add button to disable highlights when printing (it's conceivable that someone would want to print with their Hypothesis highlights, so I think it should be an option)

[...]

Issue #280, opened 2019-11-21
Hypothesis highlights appear in PDF export


dhimmel:
In https://github.com/greenelab/deep-review/pull/982#issuecomment-557165266, I noticed that a Hypothesis annotation appeared in the athenapdf PDF export of the HTML manuscript:

> ![image](https://user-images.githubusercontent.com/1117703/69357594-57053780-0c53-11ea-8fae-bb30f714dbae.png)

I believe this is occurring because we now set the rel=canonical link in the HTML `<head>`, so even though we're exporting the PDF from HTML at a local path, it's receiving the annotations from the web.

So I'm thinking we should disable hypothesis highlights for printing? @vincerubinetti what do you think?

vincerubinetti:
> I believe this is occurring because we now set the rel=canonical link in the HTML `<head>`, so even though we're exporting the PDF from HTML at a local path, it's receiving the annotations from the web.

I don't think that's the issue. This shouldn't be happening. If I recall, I added some print-specific CSS to make sure the hypothesis highlights were invisible. They may have changed their class names. I'll look into it.

vincerubinetti:
>  If I recall, I added some print-specific CSS to make sure the hypothesis highlights were invisible. 

Nevermind, apparently I didn't. I only hid the sidebar and the hypothesis highlight popup on print.

milahu:
still an issue, when i combine hypothesis with [bindery](https://github.com/evnbr/bindery)

solution: hide the custom element

```css
    @media only print {
      hypothesis-sidebar {
        display: none !important;
      }
    }
```

agitter:
Thanks for the report @milahu.  Is there a change you suggest for the Manubot rootstock `hypothesis.html` file, or does the upstream `embed.js` from Hypothesis need to be modified?

milahu:
sorry, i confused this with the hypothesis issue tracker

its an upstream issue with hypothesis

PR #345, opened 2020-06-05
fix some hypothesis issues

### LINKED ISSUE #306: Mobile UI Issue
When I open the page notes bar on the right side of the screen, the button to close the bar overlaps with the button to open the table of contents. As a result, closing the page notes/highlights bar opens the table of contents, which I then have to close to get back to reading.

Is it possible to disable the ToC button while the notes bar is expanded?

vincerubinetti:
closes https://github.com/manubot/rootstock/issues/306
closes https://github.com/manubot/rootstock/issues/293
closes https://github.com/manubot/rootstock/issues/280
closes https://github.com/manubot/rootstock/issues/281 

- changes default font sizes of h1 through h6 to be larger than base text size, breaking with browser precedent
- fixes overlap of hypothesis sidebar and table of contents plugin on small screens
- adds button to load hypothesis plugin, rather than automatically loading it

image

### DIFF
--- a/build/plugins/hypothesis.html
+++ b/build/plugins/hypothesis.html
@@ -42,9 +42,7 @@
         // create button
         const button = document.createElement('button');
         button.id = 'hypothesis_button';
-        button.innerHTML = document.querySelector(
-            '.icon_hypothesis'
-        ).innerHTML;
+        button.innerHTML = document.querySelector('.icon_hypothesis').innerHTML;
         button.title = 'Hypothesis annotations';
         button.classList.add('icon_button');
 
--- a/build/plugins/hypothesis.html
+++ b/build/plugins/hypothesis.html
@@ -63,7 +63,8 @@
         try {
             const canonical = document.querySelector('link[rel="canonical"]');
             const location = window.location;
-            const response = await fetch(query + (canonical || location).href);
+            const url = encodeURIComponent((canonical || location).href);
+            const response = await fetch(query + url);
             const json = await response.json();
             count = json.total || '-';
         } catch(error) {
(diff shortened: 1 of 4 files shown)


build/plugins/hypothesis.html
```diff
@@ -13,17 +23,83 @@
...
+    }
+
+    // insert annotations count
+    async function insertCount(button) {
+        // get annotation count from Hypothesis based on url
+        let count = '-';
+        try {
+            const canonical = document.querySelector('link[rel="canonical"]');
+            const location = window.location;
+            const response = await fetch(query + (canonical || location).href);
```
dhimmel, line 68: I'm not sure whether we need to escape the `url` / `uri` parameter.

vincerubinetti: Are you asking "do we need to escape the url parameter", because I am not doing that here. Theoretically I think the browser should escape it automatically when making the query.

vincerubinetti: If you're talking about me using `canonical`, that is so Hypothesis will return some results when you are looking at a local manuscript. Because it wont return (correct) annotation counts for `localhost:3000/`, `C:/users/vincent/test-manuscript.html`, etc.

vincerubinetti:
Looks like something's messed up with the headings. It looks fine in Chrome, but then go to print it and Heading 4 is bigger than Heading 3...

build/plugins/hypothesis.html
dhimmel: using canonical makes sense to me.

My question was whether it needs to be encoded like `encodeURIComponent(canonical)`. It's never clear to me when query string parameters need to be encoded like this.

build/plugins/hypothesis.html
```diff
@@ -13,17 +23,81 @@
...
+        document.body.append(script);
+        button.remove();
+    }
+
+    window.addEventListener('load', start);
 </script>
+
+<!-- hypothesis icon -->
+
+<template class="icon_hypothesis">
```
dhimmel, line 94: I wonder if viewers would know that the button is for Hypothes.is if we used the logo at https://simpleicons.org/icons/hypothesis.svg rather than the curly double quote

vincerubinetti: Recognition was my concern too. But I'd be fine with either. It's easy to swap out.

build/plugins/hypothesis.html
vincerubinetti: To be honest, it's not clear to me either. I think it might depend on the code reading the info on the other end too? Like does it use `decodeURIComponent`...

I think the browser should handle it in most cases though. Consider this contrived url with a space in it: `https://greenelab.github.io/xswap manuscript/`

If you do `fetch('https://greenelab.github.io/xswap manuscript/')`, it has no problem properly encoding the space (not a valid url character) and completing the request (although the request returns a 404 because it's not a real url).

If you put that url into the location bar in your browser, it also properly converts it.

dhimmel: > I think it might depend on the code reading the info on the other end too? Like does it use decodeURIComponent...

I think it probably does because elsewhere in [docs](https://h.readthedocs.io/en/latest/api-reference/v1/)

>  image

If you try the following URL, we still get 122 returned:
https://api.hypothes.is/api/search?limit=0&url=https%3A%2F%2Flelaboratoire.github.io%2Frethink-prs-ms%2F

So I think we should use `encodeURIComponent`... it's probably a bit safer by avoiding some cases where the API truncates the `url` value because of an unescaped character... although I might be missing something.

build/plugins/hypothesis.html
dhimmel: If you can get it to look nice with the Hypothesis icon, I think that'd be most recognizable for readers.

build/plugins/hypothesis.html
```diff
@@ -93,10 +94,10 @@
 
 <template class="icon_hypothesis">
     <!-- modified from: https://fontawesome.com/icons/quote-right -->
```
dhimmel, line 96:
```suggestion
    <!-- modified from: https://simpleicons.org/icons/hypothesis.svg / https://git.io/Jf1VB -->
```

vincerubinetti:
You can merge when ready.

PR #346, opened 2020-06-08
Add a space delimiter across the style

adam3smith:
Afaik, display elements don't work properly with pandoc, so I don't think the block elements will do anything for you. citeproc-js isn't happy with those two blocks in there.

### DIFF
--- a/build/assets/style.csl
+++ b/build/assets/style.csl
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-US">
   <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
@@ -50,26 +50,28 @@
   <bibliography hanging-indent="true" second-field-align="margin">
     <layout>
       <text variable="citation-number" suffix=". "/>
-      <group>
-        <text variable="title" font-weight="bold"/>
-      </group>
-      <group display="block">
-        <text macro="author"/>
-      </group>
       <group delimiter=" ">
-        <text macro="venue" font-style="italic"/>
-        <date variable="issued" prefix="(" suffix=")">
-          <date-part name="year"/>
-          <date-part name="month" form="numeric-leading-zeros" prefix="-"/>
-          <date-part name="day" form="numeric-leading-zeros" prefix="-"/>
-        </date>
-        <text variable="URL"/>
-      </group>
-      <group display="block" delimiter=" · ">
-        <text variable="DOI" text-case="lowercase" prefix="DOI: "/>
-        <text variable="PMID" prefix="PMID: "/>
-        <text variable="PMCID" prefix="PMCID: "/>
-        <text variable="ISBN" prefix="ISBN: "/>
+        <group>
+          <text variable="title" font-weight="bold"/>
+        </group>
+        <group display="block">
+          <text macro="author"/>
+        </group>
+        <group delimiter=" ">
+          <text macro="venue" font-style="italic"/>
+          <date variable="issued" prefix="(" suffix=")">
+            <date-part name="year"/>
+            <date-part name="month" form="numeric-leading-zeros" prefix="-"/>
+            <date-part name="day" form="numeric-leading-zeros" prefix="-"/>
+          </date>
+          <text variable="URL"/>
+        </group>
+        <group display="block" delimiter=" · ">
+          <text variable="DOI" text-case="lowercase" prefix="DOI: "/>
+          <text variable="PMID" prefix="PMID: "/>
+          <text variable="PMCID" prefix="PMCID: "/>
+          <text variable="ISBN" prefix="ISBN: "/>
+        </group>
       </group>
     </layout>
   </bibliography>
--- a/build/assets/style.csl
+++ b/build/assets/style.csl
@@ -47,7 +47,7 @@
       <text variable="citation-number"/>
     </layout>
   </citation>
-  <bibliography hanging-indent="true" second-field-align="margin">
+  <bibliography hanging-indent="true" second-field-align="flush">
     <layout>
       <text variable="citation-number" suffix=". "/>
       <group delimiter=" ">


rmzelle:
> I don't think the block elements will do anything for you.

https://chemical-roles.github.io/manuscript/ suggests otherwise (note the line breaks after the title and author listings).

rmzelle:
You could try this version as well: https://gist.github.com/rmzelle/a3f1fab95a4b136962fce5b1b7cdeaf8

dhimmel:
Thanks so much for the help!

In https://github.com/manubot/manubot/pull/110, I create references for all combinations of missing CSL JSON fields. Then I run a pandoc command to process a document that cites every reference.

I ran this for the following CSL style options:

```shell
# current manubot version
--csl=https://github.com/manubot/rootstock/raw/d61267c8ff1d7bc16644f4e15b814c1b4ca2bd31/build/assets/style.csl

# adam3smith version
--csl=https://github.com/manubot/rootstock/raw/7335415d96f5adc19fae0d03552a906f960468b9/build/assets/style.csl

# rmzelle version
--csl=https://gist.githubusercontent.com/rmzelle/a3f1fab95a4b136962fce5b1b7cdeaf8/raw/0e2478d17476c633b080b5197e145d1e2b858a2f/manubot.csl
```

I opened the resulting `manuscript.html` in my browser and printed to PDF, so I could attach to this issue. Here are the results:

- [manubot.pdf](https://github.com/manubot/rootstock/files/4748299/manubot.pdf)
- [adam3smith.pdf](https://github.com/manubot/rootstock/files/4748300/adam3smith.pdf)
- [rmzelle.pdf](https://github.com/manubot/rootstock/files/4748304/rmzelle.pdf)

The current manubot style lacks space when authors and venue is missing:

> image

The style by @adam3smith adds the missing space:

> image

The style by @rmzelle appears to have double new lines rather than a single newline after each group.

> image

Looking through all 256 references, I don't see any reference where the @adam3smith variant is worse than the current Manubot variant! Excellent!

For reference, here is the intended style, which works when nothing is missing:

> **This is the title**
Given-1 Family-1, Given-2 Family-2, Given-3 I. Family-3
_Container Title_ (2019-01-01) https://manubot.org
DOI: 10.0000/fake-doi

It contains four lines: title, authors, venue + date + URL, identifiers. I think the ideal behavior would be to omit a line if everything in the line is missing. Otherwise include the line. This means that there would never be any entirely bank lines, but also content from different lines would never get put on the same line.

There are still many references, where this design is violated, For reference 15 and 16 above, the title line is combined with the date/URL. When the title is missing, the initial line is blank (rather than moving the next line to the beginning):

> image

There is also a blank line when the CSL JSON contains: author, container-title, issued.

> image

Now IIRC there are some CSL/pandoc-citeproc limitations that may make the desired style impossible to achieve. @adam3smith / @rmzelle do you know if that's true? With @adam3smith's missing space fix, I think references that are blatantly wrong are addressed. So it's not a huge priority for me to address the rest, unless the fix would be straightforward.

One final note: the performance of the style using pandoc-citeproc is paramount, but it would also be nice for the style to work with other processors.

build/assets/style.csl
```diff
@@ -50,26 +50,28 @@
   <bibliography hanging-indent="true" second-field-align="margin">
     <layout>
       <text variable="citation-number" suffix=". "/>
-      <group>
-        <text variable="title" font-weight="bold"/>
-      </group>
-      <group display="block">
-        <text macro="author"/>
-      </group>
       <group delimiter=" ">
```
dhimmel, line 53: The diff here makes 7335415d96f5adc19fae0d03552a906f960468b9 look larger than it actually is.

@adam3smith is the only change that everything has been wrapped in `<group delimiter=" ">...</group>`?

build/assets/style.csl
```diff
@@ -50,26 +50,28 @@
...
-      </group>
-      <group display="block" delimiter=" · ">
-        <text variable="DOI" text-case="lowercase" prefix="DOI: "/>
-        <text variable="PMID" prefix="PMID: "/>
-        <text variable="PMCID" prefix="PMCID: "/>
-        <text variable="ISBN" prefix="ISBN: "/>
+        <group>
+          <text variable="title" font-weight="bold"/>
+        </group>
+        <group display="block">
```
dhimmel, line 57: > citeproc-js isn't happy with those two blocks in there

@adam3smith I'm guessing this comment is in reference to this line and the other one below:

```xml
        <group display="block" delimiter=" · ">
```

By not happy, do you mean that this style fails entirely with citeproc-js? Do you think the issue is with our style or citeproc-js? I.e. is this use of block valid according the spec?

rmzelle:
> The style by @rmzelle appears to have double new lines rather than a single newline after each group.

The CSL specification doesn't specify what to do with empty blocks, but this style would work better if empty blocks didn't create an empty line. You could ask if pandoc-citeproc would be willing to implement that behavior.

And I didn't notice this before, but `hanging-indent="true"` and `second-field-align="margin"` should probably be mutually exclusive, so you could try deleting either attribute.

Not sure where the double lines come from.

dhimmel:
> The CSL specification doesn't specify what to do with empty blocks, but this style would work better if empty blocks didn't create an empty line. You could ask if pandoc-citeproc would be willing to implement that behavior.

@jgm does this sound like a reasonable change for `pandoc-citeproc`? Let me know if you want me to open an issue?

dhimmel:
> And I didn't notice this before, but `hanging-indent="true"` and `second-field-align="margin"` should probably be mutually exclusive, so you could try deleting either attribute.

Based on [these docs](https://docs.citationstyles.org/en/stable/specification.html#whitespace), I think the behavior we want is `second-field-align="flush"`:

> If set, subsequent lines of bibliographic entries are aligned along the second field. With “flush”, the first field is flush with the margin.

Pandoc citeproc doesn't support this as per https://github.com/jgm/pandoc-citeproc/issues/29, but the `hanging-indent="true"` achieves something similar (but a bit less aligned). Given this situation, I'm inclined to keep both `hanging-indent` and `second-field-align` until `second-field-align` is supported in Pandoc... unless this is causing the style to fail elsewhere.

dhimmel:
@adam3smith / @rmzelle let me know if you have any additional comments here. Otherwise I'll go ahead and merge this PR, since while it doesn't fix everything, it fixes some stuff without breaking anything!

cthoyt:
@adam3smith this fixed my original issue, you can see the results in https://chemical-roles.github.io/manuscript/. Thanks so much!

Issue #344, opened 2020-06-02
Tables in docx build don't come out right


thorowitz:
The tables in my document look good in pdf but come out looking crappy in the docx file

Here is the Markdown for an example:

```markdown
| Barriers that Hinder Collaborations Between Researchers and Clinicians.|
|:------------------|
|1. Time Pressures for Clinicians.<br> - Access to clinicians is limited in terms of collaborations and participation in reader studies.<br> - Research platforms such as “pop-up” labs at national meetings should be fostered.<br> - Incentives (i.e., CEM/RME Credit) should be provided to clinicians. |
| 2. Lack of Networking Among Disciplines.<br> - Increase the probability of serendipitous collaborations by creating a "matchmaking" resource.|
| 3. Limited Access to and Utility of Image Datasets.<br> – Foster annotation projects to make datasets more useful to researchers|
```
In the pdf, it's a pretty table, with shading and borders and all the subpoenas lined up properly.

In docx, it looks like this:

Barriers that Hinder Collaborations Between Researchers and Clinicians.
--
1. Time Pressures for Clinicians. - Access to clinicians is limited in terms of collaborations and participation in reader studies. - Incentives (i.e., CEM/RME Credit) should be provided to clinicians.
2. Lack of Networking Among Disciplines. - Increase the probability of serendipitous collaborations by creating a “matchmaking” resource.
3. Limited Access to Image Datasets

Is there something I can do to improve the docx output? Maybe in the "default.docx" document?

dhimmel:
> In docx, it looks like this:

Any chance you could add a screenshot or the DOCX file?

thorowitz:
Here's the docx
<img width="652" alt="docx table" src="https://user-images.githubusercontent.com/63421291/83636503-137d7580-a574-11ea-8d6c-fab93f311cb1.png">

and here's the pdf, for reference
<img width="912" alt="pdf table" src="https://user-images.githubusercontent.com/63421291/83636542-209a6480-a574-11ea-8190-c0bf49de1a02.png">

dhimmel:
Ah, so the issue is that default DOCX style has invisible gridlines. I agree it'd be much better to see the grid-lines, and perhaps even have alternate row shading (if that's possible). I believe you can edit the DOCX style by changing [`build/themes/default.docx`](https://github.com/manubot/rootstock/blob/b7462701ff85854a67d878db4af89d0268bb11d2/build/themes/default.docx), which gets passed to pandoc:

[...]

PR #348, opened 2020-06-11
edit and add word document template(s)

vincerubinetti:
Wiki:

closes https://github.com/manubot/rootstock/issues/344

Original comment:

> image

Based on an example NIH grant given to me by @cgreene


vincerubinetti:
Will also go through the default.docx template to prune the unnecessary/unused styles and touch up the used styles.

dhimmel:
awesome! I think this will be helpful for those who want a word doc for an NIH grant. Perhaps we should change the filename mention the NIH like `nih-grant.docx`. There are many types of grants and others may require different formatting I think... @cgreene?

Can you upload the `manuscript.docx` output created using `BUILD_DOCX=true`? Would be good to see how tables and figures render.

dhimmel:
> Will also go through the default.docx template to prune the unnecessary/unused styles and touch up the used styles.

Nice. I wonder if you can resolve the invisible gridlines in tables reported in https://github.com/manubot/rootstock/issues/344

cgreene:
What is the best way to test this? I do think nih-grant.docx would be the appropriate name. It's even more specific in that it's my arbitrary NIH grant style, but it does follow their guidance so let's enforce it on others too 😁

dhimmel:
> What is the best way to test this?

Checking the output DOCX when using with rootstock delete-me.md and possibly with an actual manubot manuscript that is a grant proposal.

vincerubinetti:
@cgreene I believe you had asked to run [this document](https://github.com/greenelab/czi-rfa/blob/master/proposal.md) through the new template. Here is what it looks like:

image

[manuscript.docx](https://github.com/manubot/rootstock/files/4834218/manuscript.docx)

vincerubinetti:
[default.docx](https://github.com/manubot/rootstock/files/4834640/default.docx)

vincerubinetti:
[nih-grant.docx](https://github.com/manubot/rootstock/files/4834689/nih-grant.docx)

vincerubinetti:
After spending a lot of time on this, these two template documents are the best I can do.

I started by inputting a blank document into pandoc, which spits out a `.docx` with all of the styles it supposedly supports and understands. Here is the exact document it outputs: [reference.docx](https://github.com/manubot/rootstock/files/4834703/reference.docx)

Keeping the text content of that output basically the same, I modified all the styles to look a bit nicer, including adding borders to the tables.

There are a bunch of weird things with Word's styling for headers/tables/etc. For example, you need to have a table style (different from a regular style) called "Table", verbatim, in the template for pandoc to apply it, but Word doesn't let you create a new table style called simply "Table", because apparently it is a reserved name. However, the `reference.docx` that pandoc spits out has that table style with the correct name. Therefore, **all of the text in the template should remain there, forever**, so we don't run into weird edge cases.

There were other problems I ran into, like creating paragraph spacing and first-line indentation without affecting tables. I looked at many references online, but this ultimately boils down to stupidities in Word's design, and I can't spend too much time fiddling with them. That's why I say, the two above attached documents are the best I can do.

Here are some relevant references I looked at, in case anyone else looking at this in the future wants to tinker with it:

https://social.technet.microsoft.com/Forums/office/en-US/5d137988-9bcb-4d26-b17b-33ed2cb2c130/table-styles-paragraph-spacing-issue?forum=word
https://www.dummies.com/software/microsoft-office/word/how-to-inspect-the-current-text-style-in-word-2016/#:~:text=The%20shortcut%20key%20to%20summon,applied%20to%20your%20document's%20text.
https://github.com/jgm/pandoc/issues/3275
https://github.com/jgm/pandoc/issues/3546
https://office-watch.com/2018/make-body-text-style-appear-in-microsoft-word/
https://office-watch.com/2018/why-choose-body-text-vs-normal-in-word/
https://answers.microsoft.com/en-us/msoffice/forum/all/remove-space-after-table/1f651f2d-e630-4d53-ab1e-a3a7418faf21
https://www.techrepublic.com/blog/microsoft-office/the-right-way-to-control-spacing-in-a-word-table/
https://wordribbon.tips.net/T011729_Spacing_Before_and_After_Tables.html
https://answers.microsoft.com/en-us/msoffice/forum/all/cannot-select-positioning-in-a-new-table-style/13f242a5-973e-4b2d-9751-883470b74a14
https://answers.microsoft.com/en-us/mac/forum/macoffice2011-macword/table-paragraph-format-cant-be-changed-it-just/2affad3f-10de-485c-8a2c-3d56724c22af
https://answers.microsoft.com/en-us/msoffice/forum/msoffice_word-mso_other/i-nead-to-edit-the-format-normal-table-how-do-i-do/ab27db4e-a7c3-432f-8d3a-6c5f274f2d6a
https://www.laptopmag.com/articles/table-templates-microsoft-word
https://answers.microsoft.com/en-us/mac/forum/macoffice2011-macword/table-paragraph-format-cant-be-changed-it-just/2affad3f-10de-485c-8a2c-3d56724c22af
https://answers.microsoft.com/en-us/msoffice/forum/all/problem-with-paragraph-spacing-in-word-table/4fc381f5-b89b-4755-81db-24724ed48948

cgreene:
This is certainly close enough that one could get a good idea of page count do the final fine-tuning steps in word! Thank you for doing this!

vsmalladi:
Much better than my first pass

PR #367, opened 2020-10-02
Alternative command to push the cloned manuscript

LEskargot:
As I was struggling understanding how to "git push --set-upstream origin master" worked, I have managed to push my manuscript with another command I find more intuitive.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -70,7 +70,10 @@ Next, push your cloned manuscript:
 ```sh
 git push --set-upstream origin master
 ```
-
+An alternative way to push your cloned manuscrit would be:
+```sh
+git push --set-upstream https://github.com/OWNER/REPO
+```
 ## Continuous integration
 
 Manubot integrates with cloud services to perform continuous integration (CI).


SETUP.md
```diff
@@ -70,7 +70,10 @@ Next, push your cloned manuscript:
 ```sh
 git push --set-upstream origin master
 ```
-
+An alternative way to push your cloned manuscrit would be:
+```sh
+git push --set-upstream https://github.com/OWNER/REPO
```
dhimmel, line 75: This command is equivalent to `git push --set-upstream origin` (without `master`) assuming you used the https `set-url` command above.

Note you're missing the `$` to denote OWNER and REPO are variables.

dhimmel (commented):
In general, we want to streamline setup as much as possible, so we avoiding mentioning alternatives unless there is a functional difference. Also when things work, we tend not to suggest alternatives without extensive testing and consideration. In this case, I don't think it's worth our effort.

But thanks for getting involved and for the contribution, even if we're not able to accept this one.

SETUP.md
dhimmel: For clarification, `origin` refers to the location of the repository on GitHub. Therefore, `https://github.com/OWNER/REPO` and `origin` are synonymous. The convention is to set the remote location of the repository using a name like `origin`, and then using that to refer to that remote in commands.

Issue #368, opened 2020-10-03
is docker required for local PDF builds?


ctb:
I tried doing local builds, per https://github.com/manubot/rootstock#local-execution, and got the following error:

```
Exporting HTML manuscript
[INFO] Running filter pandoc-fignos
[INFO] Completed filter pandoc-fignos in 18 ms
[INFO] Running filter pandoc-eqnos
[INFO] Completed filter pandoc-eqnos in 10 ms
[INFO] Running filter pandoc-tablenos
[INFO] Completed filter pandoc-tablenos in 9 ms
[INFO] Running filter pandoc-manubot-cite
[INFO] Completed filter pandoc-manubot-cite in 11 ms
[INFO] Running filter pandoc-citeproc
[INFO] Completed filter pandoc-citeproc in 18 ms
Exporting PDF manuscript using Docker + Athena
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
See 'docker run --help'.
```

so, seems like it - but not mentioned anywhere in documentation. ??

dhimmel:
Currently, we check whether the docker command exists on the system. If so we use it to produce the PDF. 

https://github.com/manubot/rootstock/blob/15db6237effc6edb2b8e675ea5cb8a17188d1e2c/build/build.sh#L38-L39

https://github.com/manubot/rootstock/blob/15db6237effc6edb2b8e675ea5cb8a17188d1e2c/build/build.sh#L55-L57

When the docker command doesn't exist, we use WeasyPrint, which should be part of the conda environment:

https://github.com/manubot/rootstock/blob/15db6237effc6edb2b8e675ea5cb8a17188d1e2c/build/build.sh#L41-L44

Note that in the future we might remove WeasyPrint, but it works for now.

So the issue seems to be that your system has docker installed but not running. We should consider updating the build script to check whether docker is running.

Let me prototype a command for this.

ctb:
ah-hah! got it. thanks!

(yes, indeed, I do not keep docker running most of the time.)

dhimmel:
We could replace line 39 above with:

```bash
DOCKER_EXISTS="$(docker info >/dev/null 2>&1 || true)" 
```

You can also do this now for a local fix.

Another option is to launch Docker on macOS like https://stackoverflow.com/a/48843074/4651668, but I think I lean against it, since launching a heavy application like Docker should probably be opt-in from the user's perspective.

ctb:
[...]

PR #371, opened 2020-10-03
build.sh: skip docker when installed but not running

dhimmel:
closes https://github.com/manubot/rootstock/issues/368

https://stackoverflow.com/a/55283209/4651668
https://stackoverflow.com/a/18035947/4651668

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -35,12 +35,12 @@ pandoc --verbose \
   --defaults=common.yaml \
   --defaults=html.yaml
 
-# Return null if docker command is missing, otherwise return path to docker
-DOCKER_EXISTS="$(command -v docker || true)"
+# Set DOCKER_RUNNING to a non-empty string if docker is running, otherwise null.
+DOCKER_RUNNING="$(docker info &> /dev/null && echo "yes" || true)"
 
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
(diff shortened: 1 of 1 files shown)


build/build.sh
```diff
@@ -35,12 +35,12 @@ pandoc --verbose \
   --defaults=common.yaml \
   --defaults=html.yaml
 
-# Return null if docker command is missing, otherwise return path to docker
-DOCKER_EXISTS="$(command -v docker || true)"
+# Set DOCKER_RUNNING to a non-empty string if docker is running, otherwise null.
+DOCKER_RUNNING="$(docker info &> /dev/null && echo "yes" || true)"
```
dhimmel, line 39: Parts of this command:

1. `docker info &> /dev/null`: run docker info for exit status only, discarding both stdout and stderr
2. `&& echo "yes"`: if part 1 was a success, print "yes" to std, to set `DOCKER_RUNNING=yes`
3. `|| true` if part 1 returned a failing exit status still return a zero (success) status

agitter (approved):
Thanks for breaking down the command.  This worked as expected on my system when docker was running and not running.

dhimmel:
> This worked as expected on my system when docker was running and not running.

Which OS?

agitter:
> Which OS?

Git Bash on Windows 10

PR #372, opened 2020-10-05
change toc behavior and fix mobile click bug

### LINKED ISSUE #369: Table of contents not clickable on phones and tablets
I noticed the following unfortunate behaviour on tablets (a Galaxy tab) and mobile phones (tried on a few, including Android and IPhone): the table of contents can be toggled by clicking on the top left (so, JavaScript is activated). However it cannot be used: if I click on some document section (meaning, tap with my finger on the screen), the table of contents disappears instead of bringing me to the part of the document I clicked on.

vincerubinetti:
Closes #369 

- split auto-open/close table of contents plugin options into two separate options
- fix bug where toc panel closes when tapped on touch screen before toc link click can be registered

### DIFF
--- a/build/plugins/table-of-contents.html
+++ b/build/plugins/table-of-contents.html
@@ -1,5 +1,3 @@
-<!-- table of contents plugin -->
-
 <script>
     (function() {
         // /////////////////////////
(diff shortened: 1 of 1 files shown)


vincerubinetti:
@dhimmel @nathanael-fijalkow Here is a fix for the table of contents issue. It also changes the behavior of the plugin options.

Here is a live test:
https://jsfiddle.net/n6dswv7h/show

Please make sure this fixes the issue, and that the new behavior is okay.

dhimmel (approved):
tested on android chrome pixel 4 and on desktop chrome on linux and both worked with the default settings. I played around with startOpen and clickClose on jsfiddle. I think these new options are more intuitive.

Is there an option to not close the TOC when clicking the body (if the TOC doesn't interfere with the body)? That's the only thing missing that I could see some users wanting. But doesn't have to be part of this PR.

vincerubinetti:
We could bundle that behavior with the `clickClose` option, or we could split it up further into `startOpen`, `clickLinkClose`, `clickBodyClose`.

Or actually maybe the best thing to do is just have no option for the body click close, and only have body click close the panel if the screen is too small.

dhimmel:
> Or actually maybe the best thing to do is just have no option for the body click close, and only have body click close the panel if the screen is too small.

I like this solution. I think some users will want to have the TOC always up as they navigate through a manuscript. And if a user opens the TOC and doesn't want it, they can click the toggle button. @vincerubinetti as far as options go, your call but I think this would be a nice default. What do you think?

agitter:
> I like this solution. I think some users will want to have the TOC always up as they navigate through a manuscript. And if a user opens the TOC and doesn't want it, they can click the toggle button.

That behavior makes sense to me too.

vincerubinetti:
@dhimmel I don't have permissions to merge. This is ready to go. I implemented the last suggestion with the last `update` commit.

PR #373, opened 2020-10-05
Add a plugin for d3 loading

matthewturk:
This adds a plugin to use d3 in the document HTML.

At present, it's disabled by default, as it does two things -- it both converts figures containing SVGs into inline SVGs rather than IMG tags, and it loads d3 from a CDN.

As discussed in the documentation, it sets up a D3Loaded event that it fires once all SVG-containing figures have been inlined.

I don't know if this is worth an example, or if it even fits with the idea of Manubot.  I went this route as it seemed to be the easiest way to preserve the output from PDF and other conversions for SVG figures while still allowing some in-line interaction.

### DIFF
--- a/build/pandoc/defaults/html.yaml
+++ b/build/pandoc/defaults/html.yaml
@@ -1,5 +1,6 @@
 # Pandoc --defaults for HTML output.
 # Load on top of common defaults.
+# By default, the d3js pluging is disabled.
 to: html5
 output-file: output/manuscript.html
 include-after-body:
@@ -15,6 +16,7 @@ include-after-body:
 - build/plugins/math.html
 - build/plugins/hypothesis.html
 - build/plugins/analytics.html
+#- build/plugins/d3.html
 variables:
   math: ''
 html-math-method:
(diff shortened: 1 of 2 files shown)


agitter:
@vincerubinetti could you please comment on this plugin for adding d3 support in the HTML manuscript?

vincerubinetti:
Hi @matthewturk , I think this is a good idea, thanks for contributing.

Can you share with me any examples/references that you used to create this code. I'm unclear why certain things are being done the way they are, and there seems to be simpler ways to do the same things.

matthewturk:
Absolutely! If it could be simplified I'd be grateful for that, too.

I'll prep some examples and explanations!

vincerubinetti:
If this is from scratch, no need to prepare anything. Let me instead, then, suggest some changes and you tell me if it accomplishes the same thing or if there's something I'm missing. I'll have questions comments about specific lines in a bit.

vincerubinetti:
@matthewturk I made a new PR to replace this one, take a look. I discuss the changes there.

vincerubinetti:
Apologies for the multiple pushes.

Upon testing, this works, but it messes up a few things. The lightbox plugin wasn't designed to work on inline SVGs (which may not be a problem). Also some of the styles in the theme seem to assume an `img` tag and not a `svg` tag. Finally, this makes it so that the width/height the user sets for the image is lost, but I think I can remedy that.

matthewturk:
Ok, so I guess I should close this?

vincerubinetti:
Yes, we can incorp any changes you want to make in the other PR.

PR #374, opened 2020-10-06
GitHub Actions: use Environment Files not set-env

dhimmel:
From https://github.com/manubot/rootstock/runs/1212860980#step:2:10:

> Warning: The `set-env` command is deprecated and will be disabled soon.
> Please upgrade to using Environment Files.
> For more information see:
> https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

### DIFF
--- a/.github/workflows/manubot.yaml
+++ b/.github/workflows/manubot.yaml
@@ -17,8 +17,7 @@ jobs:
       - name: Set Environment Variables
         run: |
           TRIGGERING_SHA=${GITHUB_PULL_REQUEST_SHA:-$GITHUB_SHA}
-          TRIGGERING_SHA_7=${TRIGGERING_SHA::7}
-          echo "::set-env name=TRIGGERING_SHA_7::$TRIGGERING_SHA_7"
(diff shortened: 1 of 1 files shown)


dhimmel:
From GitHub docs:

### [Setting an environment variable](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable)

`echo "{name}={value}" >> $GITHUB_ENV`

Creates or updates an environment variable for any actions running next in a job. The action that creates or updates the environment variable does not have access to the new value, but all subsequent actions in a job will have access. Environment variables are case-sensitive and you can include punctuation.

#### [Example](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#example-4)

```
echo "action_state=yellow" >> $GITHUB_ENV
```

Running `$action_state` in a future step will now return `yellow`

agitter (approved):

PR #378, opened 2020-10-22
fix mathjax loading error

### LINKED ISSUE #377: MathJax undefined with latest plugin
Hi!  We recently updated to the current rootstock, and we're now getting "MathJax" undefined errors.  It seems that in #375 the math plugin was changed from loading:

`https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js?config=TeX-MML-AM_CHTML`

to:

`https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/config/TeX-MML-AM_CHTML.js`

Is this an intentional change?  I can't *quite* tell but reading the [mathjax docs](https://docs.mathjax.org/en/v2.7-latest/config-files.html#the-tex-mml-am-chtml-configuration-file) it seems like maybe it either should be the old style where the config is specified, or the config could be manually added to the `MathJax.Hub.Config` call in `math.html` to specify `TeX-MML-AM_CHTML`.

Or, if indeed we're misconfigured somehow in our instance of manubot, any suggestions for how to fix it on our end would be very helpful.

vincerubinetti:
closes #377

Reference: https://cdnjs.com/libraries/mathjax/2.7.5

### DIFF
--- a/build/plugins/math.html
+++ b/build/plugins/math.html
@@ -12,8 +12,8 @@
 <!-- mathjax plugin -->
 
 <script
-    src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/config/TeX-MML-AM_CHTML.js"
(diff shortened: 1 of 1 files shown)


agitter (approved):
The HTML equations from the GitHub Actions artifact look good to me:
image

Issue #381, opened 2020-10-30
Exporting manubot manuscript to Rmd


taylorreiter:
Hello! I recently used manubot to draft a collaborative document, it was a wonderful experience -- thank you for generating such a great tool! I now find need to export the manuscript to Rmarkdown. Using the output `manuscript.md`, I find with very few changes that everything knits appropriately and generates a rendered pdf of the document. However, I could not get citations to render properly. When I knit using `bibliography: references.json`, I get output like:

```
pandoc-citeproc: reference doi:10.1038/s41587-020-0439-x not found
pandoc-citeproc: reference doi:10.1371/journal.pcbi.1005755 not found
```

It seems like there is enough information output by manubot between `markdown.md`, `references.json`, and `citations.tsv` that citations/references in Rmarkdown might work relatively easily, but I couldn't figure out how to make this work. My current plan is to replace all of the ~125 citations by hand with bibtex references and generate a new bibliography, but I would love to avoid this if at all possible!

dhimmel:
Okay, I think the following code in https://github.com/manubot/rootstock/pull/382 will create a markdown file you can use with RMarkdown:

https://github.com/manubot/rootstock/blob/6645e8b722c16e9c1d78f4f35ee609114eba9228/build/build.sh#L38-L43

Let us know how that works.

PR #382, opened 2020-10-30
Output manuscript-rendered.md with pandoc markdown

dhimmel:
Helps with https://github.com/manubot/rootstock/issues/381.

Runs common filters and produces markdown output that should be portable to other Pandoc systems like Rmarkdown.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -35,6 +35,12 @@ pandoc --verbose \
   --defaults=common.yaml \
   --defaults=html.yaml
 
+echo >&2 "Exporting Markdown manuscript"
+pandoc --verbose \
+  --data-dir="$PANDOC_DATA_DIR" \
+  --defaults=common.yaml \
+  --defaults=markdown.yaml
+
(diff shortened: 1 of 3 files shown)


agitter:
Will the Markdown output be generated all the time or conditionally based on an environment variable?

dhimmel:
> Will the Markdown output be generated all the time or conditionally based on an environment variable?

I am not sure whether we should add this to rootstock at all. Currently, brainstorming all the types of markdown output users might wan. I think the two most common would be

1. `--to=markdown` (pandoc markdown after running filters, citations and references not rendered unless `--to=markdown-citations.`
2. `--to=commonmark` should render citations and bibliography.

@agitter do you think those two markdown flavors would cover most use cases?

agitter:
I think those two would cover it.  I don't recall anyone specifically needing the `--to=commonmark` version yet, but it does fill a gap.

If this is a niche and advanced use case, keeping the unmerged pull request could be a good solution so we can point to it when the question does arise.

PR #385, opened 2020-11-06
Conda environment: pin additional manubot dependencies

dhimmel:


.github/workflows/manubot.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
           restore-keys: |
             ci-cache-refs/heads/master
       - name: Install Environment
-        uses: goanpeca/setup-miniconda@v1
+        uses: conda-incubator/setup-miniconda@v1
```
dhimmel, line 35: Regarding the warnings at https://github.com/manubot/rootstock/runs/1364210024#step:5:178, we'll probably have to upgrade to `conda-incubator/setup-miniconda@v2` when that is released as per https://github.com/conda-incubator/setup-miniconda/issues/84

agitter (approved):

Issue #349, opened 2020-06-18
How will changing 'master' to 'main' affect manubot build?


jaybee84:
https://github.com/jaybee84/ml-in-rd/issues/62#issue-641308795

agitter:
Cross-referencing #341, which has some discussion and links to the `ml-in-rd` repo that show how to change the branch from which manuscripts are deployed.  Similar steps would be taken to deploy from `main`.

In terms of reconfiguring this rootstock repository to use `main` instead of `master`, my thoughts are that we may want to wait until GitHub implements its change so that our workflows and instructions are consistent with theirs.  [Recent news](https://www.cnet.com/news/microsofts-github-is-removing-coding-terms-like-master-and-slave/) makes it sound like that change is coming soon but still in progress.

We would also need to provide additional instructions in `SETUP.md` for anyone merging upstream rootsock changes.

jaybee84:
Thanks @agitter !

dhimmel:
It looks like the `master` branch name is hardcoded in several places, especially in the CI scripts and SETUP. It won't be too hard to change this repo such that new manuscripts use a different default branch. Doing this in a way that is backwards compatible for existing manuscripts that already use `master` might be more challenging.

> my thoughts are that we may want to wait until GitHub implements its change so that our workflows and instructions are consistent with theirs.

I agree. At least until GitHub provides more clarity on what the default branch name will be in the future... do we know `main` is what it will be. Hopefully, GitHub announces their plans ASAP so we can have a timeline on when we can roll this change out.

dhimmel:
More info on GitHub's updates at https://github.com/github/renaming.

Also recent blog post [GitHub Actions: Better support for alternative default branch names](https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/):

> We have updated all of our [starter workflows](https://github.com/actions/starter-workflows) to use a new `$default-branch` macro rather than the previously hardcoded `master`. You can take advantage of this feature in your [custom starter workflows as well](https://docs.github.com/en/actions/configuring-and-managing-workflows/sharing-workflow-templates-within-your-organization).

We should look into using `$default-branch` such that we begin removing hardcoded references to `master`. Ideally, we can support a default branch of any name. At that point, we might be able to switch rootstock to `main` with minimal disruption.

dhimmel:
[...]

PR #399, opened 2020-12-10
Switch default branch from master to main

dhimmel:
work in progress

refs https://github.com/manubot/rootstock/issues/349

Switch default branch from master to main, while maintaining support for existing manuscripts that continue using master branch.

TODO:

- [x] update setup docs to explain change https://github.com/manubot/manubot/pull/275
- [x] update manubot/manubot tooling that assumes master is default


ci/deploy.sh
```diff
@@ -12,7 +12,8 @@ REPO_SLUG=${TRAVIS_REPO_SLUG:-$GITHUB_REPOSITORY}
 COMMIT=${TRAVIS_COMMIT:-$GITHUB_SHA}
 CI_BUILD_WEB_URL=${CI_BUILD_WEB_URL:-$TRAVIS_BUILD_WEB_URL}
 CI_JOB_WEB_URL=${CI_JOB_WEB_URL:-$TRAVIS_JOB_WEB_URL}
-BRANCH=${TRAVIS_BRANCH:-master}
+BRANCH=${TRAVIS_BRANCH:-$DEFAULT_BRANCH}
+BRANCH=${BRANCH:-$main}
```
agitter, line 16: The approach looks good so far.  I can review when it's ready.

Did you want this here?
```suggestion
BRANCH=${BRANCH:-main}
```

dhimmel: Nice catch. Feel free to review now. Will probably add something to the setup update instructions for updating users that still have a master default. They do not need to change, but the documentation will make more sense this way.

.travis.yml
```diff
@@ -5,6 +5,7 @@ services:
   - docker
 branches:
   only:
+    - main
```
agitter, line 8: This legacy workflow could run on `main` but will only deploy when run on `master`. That's potentially inconsistent, but I don't think it is a problem.  As you note, this is a legacy workflow.

agitter (approved):
Looks good to me.  Adding a note to setup before merging is a good idea.

PR #409, opened 2021-02-02
plugin refactor

### LINKED ISSUE #241: Rename themes back to .css from .html
Even though the theme files are technically `.html` (well, they're really fragments of a valid `.html` file), I think we should rename them back to `.css`. For those who know what CSS is, they might be looking around for where to modify the CSS, and would be confused to find it in an HTML file, probably.

### LINKED ISSUE #268: Tooltips plugin malfunctions when citation hidden in <details>
@humbertodebat is leading a cool project to translate the Sci-Hub Coverage Study manuscript to spanish. We are keeping the English text but using the HTML `<details>` element to collapse it by default. [See manuscript here](https://greenelab.github.io/scihub-manuscript-es/v/353933ad77b02ebf578711ebb53f73752de55082/):

> image

For future reference, we made some changes to the CSS to style the details in https://github.com/greenelab/scihub-manuscript-es/commit/56313e8811d392cbc12f4fc6f9cc822a5aef5d2b and https://github.com/greenelab/scihub-manuscript-es/commit/3a667f27c909db1916a912a13956969ed497080e.

Because we keep the english original in `<details>` (collapsed by default), each citation appears double. When using the tooltip extension to jump between other instances of the same citation, the tooltip goes away. @vincerubinetti is there a more natural way the extension could handle this?

vincerubinetti:
Closes #268
Closes #241 

- moves repeated and shared/generic functions to separate "core" plugin
- reorganizes html.yaml config into first and third party plugins
- removes functionality to set plugin options from url
- reformats plugins with Prettier (eg 4 space tabs to 2 space)
- remove `(function () { ... } )()` anonymizer wrapper, and just make `<script>` tag into `module` to keep scope local and avoid function name conflicts. this reduces the indent of the whole script by one level.
- moves plugin specific CSS to plugins themselves

Originally I wanted all the CSS in the theme for the situation where we create another theme and need to style the plugins in that new theme too. But 1) the default theme was getting to unruly and monolithic 2) it makes more sense to group the plugin concerns together in one file 3) making a new theme has been pushed to the back burner for a long time and 4) when we do make a new theme, there are modern ways to solve the problem of styling the plugins (like CSS variables).


vincerubinetti:
@dhimmel why is the [svg image](https://raw.githubusercontent.com/manubot/resources/24ef5761233fc50e4440444e59131363ae433e35/test/vector.svg) in the `delete-me.md` being changed from `master` to `main` (and thus being a broken link).

dhimmel:
> @dhimmel why is the svg image in the delete-me.md being changed from master to main (and thus being a broken link).

Ah that happened in 690aac113b4ee7dbb32a0af10a3d36c51a207025 and was likely an accident (I must have thought this was a link to `manubot/rootstock` and not `manubot/resources`. Rather than revert that change, I am going to update the default branch for resources. I think GitHub will redirect any old links to master.

After that will review this PR! Looking forward to trying the refactored plugins.

build/pandoc/defaults/html.yaml
```diff
@@ -2,23 +2,30 @@
...
+  - build/plugins/jump-to-first.html
+  - build/plugins/lightbox.html
+  - build/plugins/link-highlight.html
+  - build/plugins/table-of-contents.html
+  - build/plugins/tooltips.html
+
+  # third-party plugins
+  - build/plugins/analytics.html
 # - build/plugins/d3.html
-include-after-body:
```
dhimmel, line 7: What is the reason for changing everything to be included before the body?

build/pandoc/defaults/html.yaml
```diff
@@ -2,23 +2,30 @@
 # Load on top of common defaults.
 to: html5
 output-file: output/manuscript.html
-# include-before-body:
+include-before-body:
+  # theme
+  - build/themes/default.html
+
```
dhimmel, line 8: slight preference to remove the blank lines in the YAML list.

Also probably best to avoid adding indentation to the list items to keep the diff simple and avoid user confusion (we've found users often mess up indentation when writing YAML.

dhimmel (commented):
net minus 700 lines... nice!

Played around with `manuscript-1.0.219-727b8ca.html` and everything looks good, but that wasn't an exhaustive test since the rootstock manuscript is simple and not all the plugins are enabled. @vincerubinetti could you render a manuscript with all plugins active. Is there a place I can see the scite plugin?

> moves repeated and shared/generic functions to separate "core" plugin

👍🏻 

> reformats plugins with Prettier

👍🏻 

> moves plugin specific CSS to plugins themselves

we can consider also closing https://github.com/manubot/rootstock/issues/241 since these files now have mixed JS/CSS?

build/pandoc/defaults/html.yaml
vincerubinetti: For our first party plugins, it doesn't matter where they go since they all wait for the document to finish loading anyway (as if they had been placed at the end). I moved them all to one command for simplicity.

build/pandoc/defaults/html.yaml
vincerubinetti: Blank lines were to clearly separate the sections of first and third party plugins. Without them, the commented out plugins also look like their own sections:
image

The indentation was just how Prettier formatted it I think. It seems [the recommended way](https://www.google.com/search?q=yaml+list+indentation) is to indent with 2 spaces. Nevertheless I've updated it to revert the indentation.

vincerubinetti:
> @vincerubinetti could you render a manuscript with all plugins active. Is there a place I can see the scite plugin?

Going to try to give you a version of the Covid manuscript with all the plugins enabled.

vincerubinetti:
[covid-19-manuscript-all-plugins.zip](https://github.com/manubot/rootstock/files/5927965/covid-19-manuscript-all-plugins.zip)

You'll need to serve the index.html from a server. Here's a preview of Scite:

image

vincerubinetti:
Sorry I uploaded the wrong file. Here's the correct one:

[covid-19-manuscript-all-plugins.zip](https://github.com/manubot/rootstock/files/5927990/covid-19-manuscript-all-plugins.zip)

You'll have to serve the index file from a server.

Here's a preview of the Scite plugin:

image

build/pandoc/defaults/html.yaml
dhimmel: Does anything have to go before body? I have a slight preference for after body since when people manually inspect the HTML file, it's more helpful to have the manuscript content near the top of the file.

build/pandoc/defaults/html.yaml
dhimmel: > Blank lines were to clearly separate the sections of first and third party plugins. Without them, the commented out plugins also look like their own sections:

Got it. The revised `###` make the sections clear to me, so I have the preference for removing blank lines.

dhimmel:
> Going to try to give you a version of the Covid manuscript with all the plugins enabled

That manuscript looks great! The scite plugin is cool. As far as the pre-existing plugins, I didn't notice any behavioral differences. Everything worked in Chrome.

build/pandoc/defaults/html.yaml
vincerubinetti: Afaik, D3 is the only problem child that needs to be before.

dhimmel: So can we keep this as `include-before-body` contains d3 and is commented out and everything else is `include-after-body`? I see why it's a bit cleaner in the config to have a single list, but would prefer to keep the existing organization since it's been working well and I prefer after-body for manual inspection of the HTML.

build/pandoc/defaults/html.yaml
```diff
@@ -3,6 +3,9 @@
 to: html5
 output-file: output/manuscript.html
 include-before-body:
```
dhimmel, line 5: Will need to comment this out as well so it's not interpreted as having a null value.

dhimmel (approved):
Awesome! Thanks @vincerubinetti for this big refactor. I'll merge now and we can address any @agitter comments in a follow up.

agitter:
> I'll merge now and we can address any @agitter comments in a follow up.

Good plan.  I want to review this but won't get to it immediately.

agitter:
I reviewed this now and don't have any additional comments.  Thanks for the big updates @vincerubinetti.  We've been using this in the COVID manuscript for the past week or so, and I haven't noticed any plugin-related issues.

PR #415, opened 2021-02-08
update scite plugin

vincerubinetti:
- removed css include (scite authors recently pushed update that includes the css in the js)
- remove vestigial `onscroll` cruft
- add js "media query" check to abort the script when printing (though it doesn't work in athena, it should work it whatever we switch to). this can prevent timeouts in contexts like appveyor where it seems to wait forever for the scite lookups to return
- add css to hide badges on print (should work in athena, but js still runs)

### DIFF
--- a/build/pandoc/defaults/html.yaml
+++ b/build/pandoc/defaults/html.yaml
@@ -23,7 +23,7 @@ include-after-body:
 - build/plugins/analytics.html
 - build/plugins/hypothesis.html
 - build/plugins/mathjax.html
-#- build/plugins/scite.html
+- build/plugins/scite.html
 variables:
   math: ""
 html-math-method:
--- a/build/pandoc/defaults/html.yaml
+++ b/build/pandoc/defaults/html.yaml
@@ -23,7 +23,7 @@ include-after-body:
 - build/plugins/analytics.html
 - build/plugins/hypothesis.html
 - build/plugins/mathjax.html
-- build/plugins/scite.html
+# - build/plugins/scite.html
 variables:
   math: ""
 html-math-method:
--- a/build/pandoc/defaults/html.yaml
+++ b/build/pandoc/defaults/html.yaml
@@ -23,7 +23,7 @@ include-after-body:
 - build/plugins/analytics.html
 - build/plugins/hypothesis.html
 - build/plugins/mathjax.html
-# - build/plugins/scite.html
+#- build/plugins/scite.html
 variables:
   math: ""
 html-math-method:
--- a/build/plugins/scite.html
+++ b/build/plugins/scite.html
@@ -14,7 +14,7 @@
 <script>
   // start script
   function start() {
-    // if printing, exit and don't show badges
+    // if printing, exit and don't run badges
     if (window.matchMedia("print").matches) return;
 
     // get citation elements
@@ -52,5 +52,11 @@
   .scite-badge {
     text-indent: 0;
     margin-top: 10px;
+  }
+
+  @media print {
+    .scite-badge {
+      display: none;
+    }
   }
 </style>
--- a/build/plugins/scite.html
+++ b/build/plugins/scite.html
@@ -5,11 +5,6 @@
   See https://scite.ai/.
 -->
 
-<link
-  rel="stylesheet"
-  type="text/css"
-  href="https://cdn.scite.ai/badge/scite-badge-latest.min.css"
-/>
 <script
   async
   type="application/javascript"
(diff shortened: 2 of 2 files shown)


cgreene:
What is the expected behavior here? I don't see scite in the PDF but I also don't see it in the HTML

vincerubinetti:
> I don't see scite in the PDF but I also don't see it in the HTML

Because scite is disabled by default

cgreene (approved):

agitter:
> Because scite is disabled by default

Can we enable it for review and then disable again before merging?

vincerubinetti:
@agitter @cgreene i've temporarily enabled scite to show that it works.

Also side note, the badge developer DM'd me the other day and said that the issue where the doi isn't found never returning has been fixed. That is, if they can't find the paper in their system, previously nothing happened and our scite badges would just be stuck on "Loading scite badge", but now they will instead just return all 0's and still show the badge.

dhimmel (approved):
looks good to me (mergeable after scite disabled by default again).

badges in HTML but not PDF makes sense to me.

vincerubinetti:
> badges in HTML but not PDF makes sense to me.

Note that this doesn't completely disable the scite plugin for printing. See this long PR discussion: https://github.com/greenelab/covid19-review/pull/846

It's an Athena bug. Should we implement the solution there here in rootstock? Idk if it's necessary because 1) scite is disabled by default 2) we might be switching off athena soon? and 3) it's only the appveyor step that has the potential to fail - the badge still gets hidden in print via the CSS but the scite javascript script itself will still run and possibly stall for appveyor (conditions of when this happen are unclear. maybe it's too many references? maybe it's references that aren't in the scite database?).

cgreene:
I think we probably should. Failing via timeout isn't terribly friendly. @agitter what do you think?

agitter:
Failing on timeouts is pretty bad.  Was that only affecting AppVeyor builds and not GitHub Actions?  Looking at some of the earlier commits in https://github.com/greenelab/covid19-review/pull/846 and @vincerubinetti's comment above, that may be the case.

I'm leaning toward not modifying the rootstock build to use the workaround we added in https://github.com/greenelab/covid19-review/pull/846.  The timeout only affects users who are using AppVeyor (i.e. not the default Manubot CI) and who enabled scite, which may be a small pool of affected users.  If we always rebuild the HTML a second time for Athena, it will slow default builds for all users.  It also creates mostly redundant HTML pandoc setting files the way we currently set it up (which could be improved).

My preference is to add a comment next to the scite plugin in `build/pandoc/defaults/html.yaml` that says something like "See https://github.com/greenelab/covid19-review/pull/846 for strategies to avoid timeouts when using scite with AppVeyor".  Then, we can continue working on migrating away from Athena as a general solution.

cgreene:
The failures were appveyor specific

vincerubinetti:
> The timeout only affects users who are using AppVeyor (i.e. not the default Manubot CI) and who enabled scite, which may be a small pool of affected users. If we always rebuild the HTML a second time for Athena, it will slow default builds for all users. It also creates mostly redundant HTML pandoc setting files the way we currently set it up (which could be improved).

I agree with this. I say keep it simple until it affects way more users.

> My preference is to add a comment next to the scite plugin

Regarding this, I want to emphasize that we're not entirely clear what is causing the problem. Casey, in the slew of different commits/workflow runs/PRs, I forget, did we actually try a version with a timeout that removes the Scite badge after ~30 sec? Did that stop AppVeyor from crashing?

It may be too many references. It may be having a particular reference that isn't found in their database. Maybe the problem won't happen anymore after their [recent update](https://github.com/manubot/rootstock/pull/415#issuecomment-776295927) to the Scite badge. Maybe my test idea to have a timeout to remove the badge doesn't even work (it just removed the HTML node for the badge, it didn't do anything to stop the Scite third-party script).

build/pandoc/defaults/html.yaml
```diff
@@ -23,7 +23,7 @@ include-after-body:
 - build/plugins/analytics.html
 - build/plugins/hypothesis.html
 - build/plugins/mathjax.html
-#- build/plugins/scite.html
+# - build/plugins/scite.html
```
agitter, line 26:
```suggestion
#- build/plugins/scite.html
```
Removing the space here so that the whitespace matches the other commented out plugins.

agitter (approved):
Looks good to me.  Based on the questions @vincerubinetti raises about not fully understanding the source of the AppVeyor timeouts, and the rarity of the AppVeyor + scite use case, I'm happy to defer any timeout workarounds until we know more.

vincerubinetti:
Feel free to merge

Issue #401, opened 2020-12-15
Create a init script? 


swamidass:
The instructions for starting a new manuscript are pretty involved: https://github.com/manubot/rootstock/blob/master/SETUP.md

Could you add a set up script which would make this work differently? E.g. make it so that manubot can be "installed," such that it provides a command line tool, which can create a new manuscript. This would reduce the process to a single command, and also make it the process maintainable.

agitter:
Initially we did not attempt to automate more of the setup steps because there were two steps that were typically performed outside the command line: creating the new repository on GitHub and setting up continuous integration.  Now that the default continuous integration moved to GitHub Actions, creating the repository is probably the only step that would need to be automated in order to create a new manuscript with a single script.

An intermediate step would be to reorder the instructions.  The user could first create the new repository at https://github.com/new.  Then, they could run a script to convert that new, empty repository into a Manubot manuscript.

Automating everything would require authenticating at the command line.  The GitHub API can be used to [create a repository](https://docs.github.com/en/free-pro-team@latest/rest/guides/getting-started-with-the-rest-api#create-a-repository).  It looks like it requires a token, which could be less user-friendly than our current setup process.  The GitHub command line interface has a [create command](https://cli.github.com/manual/gh_repo_create) and supports more [authentication options](https://cli.github.com/manual/gh_auth_login).  However, that would add an additional dependency to the setup.

PR #417, opened 2021-03-08
Setup script

nfry321:
I wanted to make a couple of manuscript repos and also wanted it to be as easy as possible to convince the rest of my team that this is the tool we should use to write papers, so I wrote a setup script. 

It takes the "Intermediate step" from #401 where the users manually create an empty repo and the script does the rest. 
This seems the most user-friendly route, even if it is not the most automated.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -32,21 +32,25 @@ Windows setup requires [Git Bash](https://gitforwindows.org/) or [Windows Subsys
 Creating a new manuscript using GitHub actions, the recommended default CI service (see below), can be achieved easily using the [setup script](https://github.com/manubot/rootstock/setup.bash).
 This simply runs the steps detailed below in the manual configuration.
 
-First, you must manually create an empty GitHub repository at <https://github.com/new>.
-Do not initialize the repository, other than optionally adding a Description.
-
-Next, run the command below.
-This will copy `setup.bash` and run it.
+Use the command below to copy `setup.bash` and run it.
 You can check the code that will be executed [here](https://github.com/manubot/rootstock/setup.bash).
 
-Replace `OWNER` and `REPO` details for your manuscript repo location:
-i.e. `https://github.com/OWNER/REPO`.
+````sh
+bash <( curl https://raw.githubusercontent.com/manubot/rootstock/main/setup.bash )
+````
+The script will then take you through the process of cloning the rootstock repo, make the changes required to use GitHub actions, edit the README to point to your repo and commit the changes.
+Your new manuscript repo is then ready for you to start adding your own content.
+
+This script does not not create the remote repo for you, so you will be prompted to manually create an empty GitHub repository at <https://github.com/new>.
+Do not initialize the repository, other than optionally adding a description.
+
+### CLI
+There is also a command line interface for users who want to create manuscripts at scale and in an automated way.
+See the help for details.
 
 ````sh
-bash <( curl https://raw.githubusercontent.com/manubot/rootstock/main/setup.bash ) -o OWNER -r REPO
+bash setup.bash --help
 ````
-The script will then clone the rootstock repo, make the changes required to use GitHub actions, edit the README to point to your repo and commit the changes.
-Your new manuscript repo is now ready for you to start adding your own content.
 
 ## Manual configuration
 
--- a/SETUP.md
+++ b/SETUP.md
@@ -36,7 +36,7 @@ Use the command below to copy `setup.bash` and run it.
 You can check the code that will be executed [here](https://github.com/manubot/rootstock/setup.bash).
 
 ````sh
-bash <( curl https://raw.githubusercontent.com/manubot/rootstock/main/setup.bash )
+bash <( curl https://github.com/manubot/rootstock/raw/main/setup.bash )
 ````
 The script will then take you through the process of cloning the rootstock repo, make the changes required to use GitHub actions, edit the README to point to your repo and commit the changes.
 Your new manuscript repo is then ready for you to start adding your own content.
--- a/SETUP.md
+++ b/SETUP.md
@@ -1,7 +1,8 @@
 # Table of contents
 
 - [Creating a new manuscript](#creating-a-new-manuscript)
-  * [Configuration](#configuration)
+  * [Using setup script](#using-setup-script)
+  * [Manual Configuration](#manual-configuration)
   * [Create repository](#create-repository)
   * [Continuous integration](#continuous-integration)
     + [GitHub Actions](#github-actions)
@@ -27,7 +28,26 @@ These steps should be performed in a command-line shell (terminal), starting in 
 Setup is supported on Linux, macOS, and Windows.
 Windows setup requires [Git Bash](https://gitforwindows.org/) or [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/faq).
 
-## Configuration
+## Using setup script
+Creating a new manuscript using GitHub actions, the recommended default CI service (see below), can be achieved easily using the [setup script]((https://github.com/manubot/rootstock/setup.bash). This simply runs the steps detailed below in the Manual Configuration.
+
+First, you must manually create an empty GitHub repository at <https://github.com/new>.
+Do not initialize the repository, other than optionally adding a Description.
+
+Next, copy `setup.bash` to the directory where you would like to clone your manuscript repo, then make it executable.
+
+Then run the script using the `OWNER` and `REPO` details for your manuscript repo location:
+i.e. `https://github.com/OWNER/REPO`.
+
+````sh
+# Make script executable
+chmod +x setup.bash
+# Run script with your repo details as arguments
+./setup.bash -o OWNER -r REPO
+````
+The script will then clone the rootstock repo, make the changes required to use GitHub actions, edit the README to point to your repo and commit the changes.
+
+## Manual Configuration
 
 First, you must configure two environment variables (`OWNER` and `REPO`).
 These variables specify the GitHub repository for the manuscript (i.e. `https://github.com/OWNER/REPO`).
--- a/SETUP.md
+++ b/SETUP.md
@@ -45,6 +45,7 @@ bash <( curl https://raw.githubusercontent.com/manubot/rootstock/main/setup.bash
 ````
 The script will then clone the rootstock repo, make the changes required to use GitHub actions, edit the README to point to your repo and commit the changes.
 The repo is now ready for you to start adding your own content.
+
 ## Manual Configuration
 
 First, you must configure two environment variables (`OWNER` and `REPO`).
--- a/SETUP.md
+++ b/SETUP.md
@@ -34,26 +34,17 @@ Creating a new manuscript using GitHub actions, the recommended default CI servi
 First, you must manually create an empty GitHub repository at <https://github.com/new>.
 Do not initialize the repository, other than optionally adding a Description.
 
-Next, copy `setup.bash` to the directory where you would like to clone your manuscript repo, then make it executable.
+Next, run ths command below. This will copy `setup.bash` and run it.
+You can check the code that will be executed [here](https://github.com/manubot/rootstock/setup.bash).
 
-Then run the script using the `OWNER` and `REPO` details for your manuscript repo location:
+Replace `OWNER` and `REPO` details for your manuscript repo location:
 i.e. `https://github.com/OWNER/REPO`.
 
 ````sh
-# create file in dir to contain the manuscript repo
-touch setup.bash
-
-# Use your favourite editor to copy and paste in the
-# contents of https://github.com/manubot/rootstock/setup.bash
-
-# Make script executable
-chmod +x setup.bash
-
-# Run script with your repo details as arguments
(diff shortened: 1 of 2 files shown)


SETUP.md
```diff
@@ -27,7 +28,33 @@ These steps should be performed in a command-line shell (terminal), starting in
...
+
+````sh
+# create file in dir to contain the manuscript repo
+touch setup.bash
+
+# Use your favourite editor to copy and paste in the
+# contents of https://github.com/manubot/rootstock/setup.bash
+
+# Make script executable
+chmod +x setup.bash
```
dhimmel, line 50: you probably considered:

```bash
curl URL_TO_setup.bash | bash
```

This would be more convenient, but perhaps less portable and doesn't give the user the chance to audit the code.

setup.bash
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+# Setup Manubot
+# Based on https://github.com/manubot/rootstock/blob/main/SETUP.md
+# This is designed to be run from the terminal on linux
+
+# Stop on first error.
+set -e
+
+usage() { echo "Usage: $0 [-o <owner-string>] [-r <repo-string>]" 1>&2; exit 1; }
```
dhimmel, line 9: in general we use long args for readability. Would it be possible to support --owner and --repo (or whatever we'd call the long argument)?

dhimmel (commented):
Really awesome! Thanks for opening a PR.

I'm looking forward to reviewing in more depth and testing out the script. I imagine others like @agitter will want to provide feedback as well.

SETUP.md
nfry321: TBH I didn't think about this as I'm relatively new to linux!
This is much neater so I will change the instructions to use curl, syntax has to be a bit different to allow user interaction once the script starts.

setup.bash
nfry321: Done, changed from using `optargs` to `optarg` which is not built into bash but should come as standard on linux (I believe).

setup.bash
```diff
@@ -0,0 +1,124 @@
...
+git clone --single-branch https://github.com/manubot/rootstock.git ${REPO}
+cd ${REPO}
+
+echo
+echo "Setup tracking using https..."
+echo
+# Configure remotes
+git remote add rootstock https://github.com/manubot/rootstock.git
+# Option A: Set origin URL using its web address
+git remote set-url origin https://github.com/${OWNER}/${REPO}.git
```
dhimmel, line 95: The lack of support for ssh keys is problematic. If you have 2FA enabled on GitHub, you [can't authenticate](https://docs.github.com/en/github/authenticating-to-github/accessing-github-using-two-factor-authentication#using-two-factor-authentication-with-the-command-line) with your username/password in git.

I wonder if we [could test](https://docs.github.com/en/github/authenticating-to-github/testing-your-ssh-connection) whether SSH access to github is configured and if so, use an ssh conneciton.

dhimmel (commented):
Looking great! I have quite a few comments, but I want to respectful of your time, so let us know whether you are running out of contribution time.

Tested via the following (got stuck on authentication as mentioned inline):

```bash
bash <( curl -L https://github.com/manubot/rootstock/raw/89cead81531d694aee796c07d78e710f7740ecd4/setup.bash ) --owner dhimmel --repo rootstock-pr-417-test
```

SETUP.md
dhimmel: > I didn't think about this as I'm relatively new to linux!

A small note here, this command should not be limited to working on just linux. Ideally it will work on macOS, linux, and Windows Bash shells. The actual setup script can require bash, but users will likely enter the initial command from some non-bash shells.

> TBH I didn't think about this as I'm relatively new to linux!

Welcome! I've been using linux for decades, but still don't know many of the shell script constructs and syntaxes you've masterfully employed. haha

setup.bash
dhimmel: Looks like getopt is also on macOS: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/getopt.3.html

SETUP.md
```diff
@@ -27,7 +28,25 @@ These steps should be performed in a command-line shell (terminal), starting in
...
+Do not initialize the repository, other than optionally adding a Description.
+
+Next, run ths command below. This will copy `setup.bash` and run it.
+You can check the code that will be executed [here](https://github.com/manubot/rootstock/setup.bash).
+
+Replace `OWNER` and `REPO` details for your manuscript repo location:
+i.e. `https://github.com/OWNER/REPO`.
+
+````sh
+bash <( curl https://raw.githubusercontent.com/manubot/rootstock/main/setup.bash ) -o OWNER -r REPO
```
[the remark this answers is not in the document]
agitter:
```
bash <( curl --location https://github.com/manubot/rootstock/raw/89cead81531d694aee796c07d78e710f7740ecd4/setup.bash )
```
worked on Git Bash (by printing the usage message)

agitter (commented):
This is fantastic, thanks @nfry321.  @dhimmel has already provided thoughtful comments about the setup script, so I focused on the instructions and testing it in Git Bash for Windows.

The setup script ran as expected in Git Bash, and the `getopt` command is available.  I was able to create https://github.com/agitter/rootstock-pr-417.  I did not test all possible errors.

SETUP.md
nfry321: Haha thanks. 

> I want to respectful of your time, so let us know whether you are running out of contribution time.

I'm happy to continue working on this but it maybe a bit slow depending on how much I've got on that's all.

setup.bash
```diff
@@ -0,0 +1,124 @@
...
+     echo "Continuing Setup..."
+     echo
+     break
+     ;;
+  [nN][oO]|[nN])
+     echo
+     echo "Go to https://github.com/new and create https://github.com/${OWNER}/${REPO}"
+     echo "Note: the new repo must be completely empty or the script will fail."
+     echo
+     exit 1
```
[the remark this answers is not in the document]
nfry321: Yes, good idea

setup.bash
```diff
@@ -0,0 +1,124 @@
...
+git push --set-upstream origin main
+
+# To use GitHub Actions only:
+echo "Setup for GitHub Actions ONLY..."
+# remove Travis CI config
+git rm .travis.yml
+# remove AppVeyor config
+git rm .appveyor.yml
+# remove ci/install.sh if using neither Travis CI nor AppVeyor
+git rm ci/install.sh
```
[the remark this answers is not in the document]
nfry321: I'd lean towards removing it, but that is on the assumption that people tend not to do upgrades unless they really need to.

setup.bash
```diff
@@ -0,0 +1,124 @@
...
+fi
+
+# Check Remote Repo Exists.
+echo
+while true
+do
+ read -r -p "Have you manually created https://github.com/${OWNER}/${REPO}? [y/n] " input
+
+ case $input in
+   [yY][eE][sS]|[yY])
```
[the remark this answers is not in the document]
nfry321: Fair point, I have made the process more interactive so it can be run without any arguments as you suggested, and also with the option for no interaction using a --yes flag.

setup.bash
nfry321: I have added an option, both interactive and CLI, for using ssh. 

I'm not sure about testing for SSH as it seems more intrusive with them having to 

> authenticate this action using your password, which is the SSH key passphrase

SETUP.md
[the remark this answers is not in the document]
nfry321: [This](https://ss64.com/osx/syntax-redirection.html) suggests that process substitution will work on macOS, but I can't test. 
I'll update the url format.

setup.bash
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
nfry321: Not sure, I just used `.bash` as a convention I picked up from others I work with. 
There is a [third option](https://stackoverflow.com/a/27824204) of no extension at all, but personally I like the verbosity of `.bash`.

setup.bash
```diff
@@ -0,0 +1,230 @@
...
+sed "s/manubot\.github\.io\/rootstock/${OWNER}\.github\.io\/${REPO}/g" README.md > tmp && mv -f tmp README.md
+
+echo "Committing rebrand..."
+git add --update
+git commit --message "Brand repo to $OWNER/$REPO"
+git push origin main
+echo
+echo "Setup complete"
+echo
+echo "The new repo as been created here: $(pwd)/${REPO}"
```
dhimmel, line 227:
```suggestion
echo "The new repo has been created at $(pwd)"
```

working dir is repo

setup.bash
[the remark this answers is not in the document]
dhimmel: Interesting. I thought the extension was important for syntax highlighting, but vscode can look at the shebang to decipher syntax. Not sure about GitHub.

The extension is nice IMO because it:

1. makes it easier to filter for bash scripts based on filename
2. makes it clear it needs to be used with bash (and we're not executing the file directly)
3. allows for a directory without the extension.

I am okay with keeping .bash and then switching our other scripts to use .bash (if they require bash).

dhimmel (commented):
two minor tweaks and then I'll approve. Will also wait for @agitter to approve.

dhimmel (approved):
Thanks. Writing these bash scripts is hard.

agitter:
I'll take a final pass this weekend.  Based on the recent iterations, I don't expect to have any new suggestions.

SETUP.md
```diff
@@ -27,7 +28,31 @@ These steps should be performed in a command-line shell (terminal), starting in
 Setup is supported on Linux, macOS, and Windows.
 Windows setup requires [Git Bash](https://gitforwindows.org/) or [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/faq).
 
-## Configuration
+## Using setup script
+Creating a new manuscript using GitHub actions, the recommended default CI service (see below), can be achieved easily using the [setup script](https://github.com/manubot/rootstock/setup.bash).
```
agitter, line 32:
```suggestion
Creating a new manuscript using GitHub actions, the recommended default CI service (see below), can be achieved easily using the [setup script](https://github.com/manubot/rootstock/blob/main/setup.bash).
```

SETUP.md
```diff
@@ -27,7 +28,31 @@ These steps should be performed in a command-line shell (terminal), starting in
...
+This simply runs the steps detailed below in the manual configuration.
+
+Use the command below to copy `setup.bash` and run it.
+You can check the code that will be executed [here](https://github.com/manubot/rootstock/setup.bash).
```
agitter, line 36:
```suggestion
You can check the code that will be executed [here](https://github.com/manubot/rootstock/blob/main/setup.bash).
```

SETUP.md
```diff
@@ -27,7 +28,31 @@ These steps should be performed in a command-line shell (terminal), starting in
...
+There is also a command line interface for users who want to create manuscripts at scale and in an automated way.
+See the help for details.
+
+````sh
+bash setup.bash --help
+````
+
+## Manual configuration
 
 First, you must configure two environment variables (`OWNER` and `REPO`).
```
agitter, line 57:
```suggestion
If you do not wish to use the above setup script to configure your new manuscript repository, you can instead execute the steps manually.
First, you must configure two environment variables (`OWNER` and `REPO`).
```

Trying to make it more clear that a user only needs to follow either the setup script or the manual configuration instructions.

setup.bash
```diff
@@ -0,0 +1,230 @@
+#!/bin/bash
+# Setup Manubot
+# Based on https://github.com/manubot/rootstock/blob/main/SETUP.md
+# This is designed to be run from the terminal on linux
```
agitter, line 4:
```suggestion
# This is designed to be run from the bash terminal
```

It works in bash on Windows as well, and will probably work in macOS.

setup.bash
```diff
@@ -0,0 +1,230 @@
...
+
+# Configure remotes
+git remote add rootstock https://github.com/manubot/rootstock.git
+
+# Check auth method
+if [[ "$YES" == '0' ]]; then
+  while true
+  do
+   echo
+   read -r -p "Would you like to use SSH to authenticate your GitHub account? [y/n]" input
```
agitter, line 168:
```suggestion
   read -r -p "Would you like to use SSH to authenticate your GitHub account? [y/n] " input
```

agitter (changes requested):
I did more testing of various failure modes, and everything worked expected.  This is excellent.

I have a few final small suggestions and then am ready to merge.

If we want to make this even more novice-friendly, we could make a follow up pull request to document the common reasons the script fails and how to interpret the git error messages like "remote: Repository not found." and "error: failed to push some refs to...".

nfry321:
Thanks for the suggestions to improve this @agitter & @dhimmel, it's definitely more user friendly now. 
It looks like the outstanding things are just improving the documentation further which I think would be best left to another PR.

agitter (approved):
> It looks like the outstanding things are just improving the documentation further which I think would be best left to another PR.

I agree.  The script is ready to merge.  Thanks a lot for this contribution.

PR #426, opened 2021-05-29
Build docker image via GitHub Actions

dhimmel:
Very preliminary proof of concept. The idea is to have CI create a Docker image for the environment. This would have two potential benefits:

1. improve build times for subsequent builds that do not modify the environment
2. make the CI environment available for local usage, allowing users to return to any past environment


agitter:
Great idea.  Both of those benefits sound good to me and should be achievable.

How much of the CI environment will eventually be added into the Docker image?  Initially it has the conda environment.  Will spellcheck, PDF generation, or anything else be added later?  In particular, I see that the current version uses weasyprint.  I don't know much about running Docker (for athenapdf) inside Docker but presume that isn't recommended.

If you don't want to use the older style of conda activation (add to PATH and `source activate`), we have used the template from this [blog post](https://pythonspeed.com/articles/activate-conda-dockerfile/) in other projects.  It sets the `ENTRYPOINT` to use `conda run` to run commands inside the desired environment.

PR #425, opened 2021-05-29
upgrade pandoc to version 2.14

dhimmel:
supersedes https://github.com/manubot/rootstock/pull/386

PR #427, opened 2021-06-02
Upgrade pandoc to 2.14 and fix plugin breaks

vincerubinetti:
supersedes #425


build/plugins/anchors.html
```diff
@@ -10,7 +10,8 @@
 <script type="module">
   // which types of elements to add anchors next to, in "document.querySelector"
   // format
-  const typesQuery = 'h1, h2, h3, [id^="fig:"], [id^="tbl:"], [id^="eq:"]';
+  const typesQuery =
+    'h1, h2, h3, div[id^="fig:"], div[id^="tbl:"], span[id^="eq:"]';
```
dhimmel, line 14: Is `div[id^="fig:"]`the workaround for double table ids (on the div and table) mentioned at https://github.com/manubot/rootstock/pull/425#issuecomment-853395226? It works by only matching div ids that start with `tbl:` as opposed to all elements?

Just to confirm, it is okay for us to proceed despite the duplicate ids, because this PR handles it?

build/themes/default.html
```diff
@@ -495,6 +494,15 @@
...
   }
 
+  /* -------------------------------------------------- */
+  /* references */
+  /* -------------------------------------------------- */
+
+  .csl-entry {
+    margin-top: 20px;
+    margin-bottom: 20px;
+  }
```
dhimmel, line 504: I think I'd be in favor of a slightly smaller margin between references than:

image

I think we do want some, but since reference sections can get quite long, perhaps we could get by with slightly less? Maybe 15 (or 50% to 75% of the current spacing)?

dhimmel (commented):
Reviewing https://ci.appveyor.com/api/buildjobs/992931201tet4bko/artifacts/manuscript-1.0.244-b5233d9.html and everything looks good (one small comment about references spacing).

Thanks a lot!

build/plugins/anchors.html
dhimmel: Opened https://github.com/tomduck/pandoc-tablenos/issues/31

vincerubinetti: > Is div[id^="fig:"]the workaround for double table ids (on the div and table) mentioned at #425 (comment)? It works by only matching div ids that start with tbl: as opposed to all elements?

That's correct. I've essentially made the queries more specific. As long as pandoc/table-nos is consistent and never outputs a table without the div wrapper, this should be fine.

Looked at the issue, 👍 FWIW id's being unique isn't just a suggestion, it's a requirement of the spec and can cause unexpected behavior. It shouldn't cause anything to crash though.

build/themes/default.html
vincerubinetti: Reduced to 15px. I wouldn't recommend going lower than that, as it becomes difficult to distinguish for sighted and hard-of-seeing people alike.

build/plugins/anchors.html
dhimmel: Got it.

> As long as pandoc/table-nos is consistent and never outputs a table without the div wrapper, this should be fine.

Based on the early discussion in the issue, it seems like the most likely fix is that pandoc-tablenos stops adding the div wrapper for tables since the id is now set in `<figure>`.

So if we upgrade pandoc-tablenos, we might have to tweak this, but that's okay. Let's wait to see what solution pandoc-tablenos adopts and merge this now.

build/themes/default.html
dhimmel: Here's 15. Looks good:

image

build/plugins/anchors.html
vincerubinetti: Oh if that’s the case we’ll need to re-add the plug-in to wrap tables in divs for scrolling purposes 😞

dhimmel (approved):
LGTM. Will wait for @agitter  to review before merging

dhimmel:
@agitter following up in case you'd like to review

USAGE.md
```diff
@@ -133,14 +133,23 @@ Note that multiple citations must be semicolon separated.
...
 
-Citation keys must adhere to the syntax described in the [Pandoc manual](https://pandoc.org/MANUAL.html#citations):
-
-> The citation key must begin with a letter, digit, or `_`, and may contain alphanumerics, `_`, and internal punctuation characters (`:.#$%&-+?<>~/`).
-
-To evaluate whether a citation key fully matches this syntax, try [this online regex](https://regex101.com/r/mXZyY2/latest).
-If the citation key is not valid, use the [citation aliases](#citation-aliases) workaround below.
-This is required for citation keys that contain forbidden characters such as `;` or `=` or end with a non-alphanumeric character such as `/`.
-<!-- See [jgm/pandoc#6026](https://github.com/jgm/pandoc/issues/6026) for progress on a more flexible Markdown citation key syntax. -->
+The citation key syntax is described in the [Pandoc manual](https://pandoc.org/MANUAL.html#citations):
```
agitter, line 136:
```suggestion
The citation key syntax is described in the [Pandoc manual](https://pandoc.org/MANUAL.html#citation-syntax):
```

Is this part of the manual more relevant now?

build/environment.yml
```diff
@@ -21,14 +20,15 @@ dependencies:
...
     - opentimestamps==0.4.1
-    - pandoc-eqnos==2.3.0
-    - pandoc-fignos==2.3.1
-    - pandoc-tablenos==2.2.2
-    - pandoc-xnos==2.4.3
+    - pandoc-eqnos==2.5.0
+    - pandoc-fignos==2.4.0
+    - pandoc-tablenos==2.3.0
+    - pandoc-xnos==2.5.0
+    - panflute==2.1.0
```
agitter, line 31: Do we want to get this from conda?  2.1.0 is now available through conda-forge: https://anaconda.org/conda-forge/panflute

agitter (approved):
Looks good to me.  I have two minor suggestions, both optional.

agitter:
The builds are failing because adding panflute to the conda-installed packages revealed an actual conflict.  panflute 2.1 [supports](https://github.com/sergiocorreia/panflute#supported-pandoc-versions) pandoc versions 2.11.0.4—2.13.x.  When we install panflute from pip, it can't check the pandoc version.  The conda-forge recipe does check the pandoc version.

Our environment appears to work based on the HTML artifacts above when we ran with panflute 2.1.0 and pandoc 2.14.0.1.  Should we revert 9ae8d19a91801713c8a85b70176945024c7dc485 and proceed to merge?

dhimmel:
Thanks all for the help! Especially @vincerubinetti for the frontend edits. The diff looks simple in the end, but this includes some major upgrades in our dependencies.

Issue #437, opened 2021-09-02
Giant orcid logos in docx output


ch728:
My docx output has giant  orcid logos in  it.  I do not have this problem in the html or pdf output.
![Screenshot from 2021-09-02 16-47-01](https://user-images.githubusercontent.com/36375878/131917468-56ac24cc-b07a-4e0d-b776-38e09d8b97ae.png)

agitter:
The issue is that these are svg images and docx readers (especially older versions of Word) do not have great svg support.  There is some historical discussion of that in https://github.com/manubot/rootstock/pull/40.  This is a general pandoc issue, and there was an update within the past week to improve svg docx support https://github.com/jgm/pandoc/issues/4058.

I tested your manuscript.  Word 2016 on Windows doesn't render the svg images at all:
![image](https://user-images.githubusercontent.com/3532898/132054286-0925a341-f518-4e26-9691-f8fd1ed34c15.png)

Some users have modified `content/00.front-matter.md` to remove these images or customize the build script to exclude the images when converting to docx.  Others manually delete them from the docx.  We don't have a good general solution yet.  The next step may be to test the next pandoc release to explore its updated support.

twrightsman:
This can be fixed on the rootstock side by manually defining the size of the SVG images in `content/00.front-matter.md`. I used 16px because that's the size defined in the SVG files.

```
diff --git a/content/00.front-matter.md b/content/00.front-matter.md
index 2f01ad2..a9c1b5c 100644
--- a/content/00.front-matter.md
+++ b/content/00.front-matter.md
@@ -25,15 +25,15 @@ on {{manubot.date}}.
 {% for author in manubot.authors %}
 + **{{author.name}}**<br>
   {%- if author.orcid is defined and author.orcid is not none %}
-    ![ORCID icon](images/orcid.svg){.inline_icon}
+    ![ORCID icon](images/orcid.svg){.inline_icon width=16 height=16}
     [{{author.orcid}}](https://orcid.org/{{author.orcid}})
   {%- endif %}
   {%- if author.github is defined and author.github is not none %}
-    · ![GitHub icon](images/github.svg){.inline_icon}
+    · ![GitHub icon](images/github.svg){.inline_icon width=16 height=16}
     [{{author.github}}](https://github.com/{{author.github}})
   {%- endif %}
   {%- if author.twitter is defined and author.twitter is not none %}
-    · ![Twitter icon](images/twitter.svg){.inline_icon}
[...]

PR #439, opened 2021-09-29
Explicitly define front matter icon dimensions

twrightsman:
closes https://github.com/manubot/rootstock/issues/437


dhimmel (approved):
Thanks!

PR #443, opened 2021-11-03
Upgrade environment on 2021-11-02

dhimmel:


agitter:
I see in the [docs](https://github.com/conda-incubator/setup-miniconda#example-6-mamba) that using mamba requires including conda-forge as part of the channels.  Do we know how this interacts with the channels listed in the .yml file?  Are those ignored now that channels are specified in the setup action?

Checking the latest build in this pull request and the latest on the main branch, the environment setup time drops from ~2 min to ~1.5 min.

dhimmel:
> the environment setup time drops from ~2 min to ~1.5 min.

I had switched to mamba here since I wanted a better error message about which dependency was causing the unresolvable errors. But given the reduced runtime, it's something we could consider switching to.

I imagine the environment channels override the mamba-default channels:

https://github.com/manubot/rootstock/blob/0b426ebdaa751d5baff32f43f53f9d12ad0ac26b/build/environment.yml#L2-L3

We should see if we can remove `defaults` from `channels`, since our environment only uses conda-forge. Apparently the `channel-priority: true` [setting](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html#strict-channel-priority) might speed things up as well.

agitter:
> We should see if we can remove defaults from channels, since our environment only uses conda-forge.

That's a good idea.  Let's try that before merging.

dhimmel:
Closing this PR since it's superseded by https://github.com/manubot/rootstock/pull/448. Might later open another PR to use mamba for faster environment installation on CI builds.

Issue #406, opened 2021-01-01
Supporting local Windows weasyprint builds


agitter:
The Manubot build [environment](https://github.com/manubot/rootstock/tree/main/build#environment) is not currently supported on Windows.  However, the main limitation was previously the [pango package](https://anaconda.org/conda-forge/pango), and pango now has win-64 support on the conda-forge channel. 

The [current conda environment](https://github.com/manubot/rootstock/blob/2191d929a12a233347ee79c673b8f2971c10e6dd/build/environment.yml) cannot be installed on Windows and gives the ResolvePackageNotFound error reported in #405.  However, if I unpin the cairo, cairocffi, and pango versions, I am able to install the manubot conda environment.  I end up with
- cairo 1.16.0
- cairocffi 1.2.0
- pango 1.42.4

`bash build/build.sh` shows they are still missing dependencies:
```
OSError: cannot load library 'gobject-2.0': error 0x7e.  Additionally, ctypes.util.find_library() did not manage to locate a library called 'gobject-2.0'
```
This is related to https://github.com/Kozea/WeasyPrint/issues/971.

I tried `conda install -c conda-forge gtk3` and installed version 3.24.24.  That did not resolve the error.

It looked like I had the missing DLLs in `.conda\envs\manubot\Library\bin` but they did not have the expected names.  I copied `gobject-2.0-0.dll` as `libgobject-2.0-0.dll`, `pango-1.0-0.dll` as `pango-1.0.dll`, and `pangocairo-1.0-0.dll` as `pangocairo-1.0.dll`.  This gave a new error further into the weasyprint build:
```
(process:11000): GLib-GObject-WARNING **: 08:49:54.950: cannot register existing type 'PangoLayout'
(process:11000): GLib-CRITICAL **: 08:49:54.950: g_once_init_leave: assertion 'result != 0' failed
(process:11000): Pango-CRITICAL **: 08:49:54.950: pango_layout_get_iter: assertion 'PANGO_IS_LAYOUT (layout)' failed
Error producing PDF.
```

Perhaps local Windows builds with weasyprint are now possible, but I'm not planning to troubleshoot this further at this time.

agitter:
I confirmed that building the manuscript locally on Windows with Git for Windows does work when Docker is available.  This line in the build script
https://github.com/manubot/rootstock/blob/2191d929a12a233347ee79c673b8f2971c10e6dd/build/build.sh#L69

needs to be changed to:
```sh
[...]

PR #448, opened 2021-11-29
Upgrade environment fully on 2021-11-29

dhimmel:
In the past we've had challenges trying to upgrade package versions one-by-one due to unresolvable conflicts. Therefore, I'm trying this new strategy to let conda set all the versions:

```bash
# remove existing conda environment
conda env remove --name manubot

# manually edit environment.yml to remove all pineed versions

# install environment letting conda resolve all versions
conda env create --file build/environment.yml

# Export full environment which includes all packages, versions, and build numbers (on Linux)
conda env export --name=manubot > build/environment-full.yml
```

The downside of this strategy is it will break if there were packages that we were intentionally holding back. I know we were holding weasyprint or its dependencies back, I think for Windows compatibility. 

TODO:

- [x] remove `build/environment-full.yml`
- [x] test local linux build using weasyprint: [manuscript.pdf](https://github.com/manubot/rootstock/files/7621673/manuscript.pdf)
- [x] test if the updated environment installs on windows
- [ ] test local windows build using weasyprint


dhimmel:
Note the following from [build/README.md](https://github.com/manubot/rootstock/blob/8086a3985300c203353acbe44a2fc908f677d563/build/README.md?plain=1#L44-L45):

> The environment should successfully install on both Linux and macOS. However, it will fail on Windows due to the pango dependency.

We should remove this if the new environment now works on Windows.

dhimmel:
This environment is building the manuscript without docker on linux and macos via GitHub actions! Windows builds work up to the WeasyPrint export where they hit https://github.com/conda-forge/weasyprint-feedstock/issues/31.

agitter:
This looks like a smarter way to update dependencies without chasing down incompatible versions and having to fix them individually.  This latest environment gets closer to functional Windows builds than my previous attempt in #406.

When I test the environment file from 18091c8 locally in Windows, fontconfig 2.13.1 is supposedly installed.  I get the same error shown in GitHub Actions.  I'll follow up in the weasyprint issue.

If we're updating many package versions or major packages like pandoc, it would be nice to have a way to do regression testing.

dhimmel:
> If we're updating many package versions or major packages like pandoc, it would be nice to have a way to do regression testing.

Yes very! But not sure how we'd make these tests.

agitter:
It's tricky with this repository serving as a template.  We don't want to add tests here.

A better solution may be to have a separate repository that automatically tracks branches in the rootstock repository and runs CI tests with the updated rootstock.  I haven't set up a cross-repository workflow before, so I don't have a specific implementation in mind.

Issue #440, opened 2021-10-19
Conditionally disable the docker builds


multimeric:
I'm on a machine with Docker installed, so `build/build.sh` will use it to build the manuscript. However, this is quite slow for me, and I prefer the WeazyPrint method 100% of the time when I'm drafting. For this reason I would appreciate a way to turn off the docker builds by _choice_. I know that the build shell script is part of the repo and therefore editable, so I have done this:

```patch
diff --git a/build/build.sh b/build/build.sh
index 778c958..5b4c39b 100755
--- a/build/build.sh
+++ b/build/build.sh
@@ -49,3 +49,3 @@ DOCKER_RUNNING="$(docker info &> /dev/null && echo "yes" || true)"
 # If Docker is not available, use WeasyPrint to create PDF
-if [ "${BUILD_PDF,,}" != "false" ] && [ -z "$DOCKER_RUNNING" ]; then
+if [[ "${BUILD_PDF,,}" != "false" && (-z "$DOCKER_RUNNING" || -n "$SKIP_DOCKER") ]]; then
   echo >&2 "Exporting PDF manuscript using WeasyPrint"
@@ -62,3 +62,3 @@ fi
 # If Docker is available, use athenapdf to create PDF
-if [ "${BUILD_PDF,,}" != "false" ] && [ -n "$DOCKER_RUNNING" ]; then
+if [[ "${BUILD_PDF,,}" != "false" && -n "$DOCKER_RUNNING" && -z "$SKIP_DOCKER" ]]; then
   echo >&2 "Exporting PDF manuscript using Docker + Athena"
```
(in hindsight it would probably have been neater to just add this logic to `DOCKER_RUNNING`)

I'm wondering if there's an interest in pushing this feature upstream.

dhimmel:
For option names, I don't have a strong preference between `SKIP_DOCKER`, `USE_DOCKER`, `MANUBOT_SKIP_DOCKER` and `MANUBOT_USE_DOCKER`. We did want to start using a manubot prefix `MANUBOT_` for manubot related options, but this is also a bit inconsistent at this point.

dhimmel:
@multimeric if you have time, would love your feedback on https://github.com/manubot/rootstock/pull/449?

PR #449, opened 2021-11-29
build: MANUBOT_USE_DOCKER environment variable

dhimmel:
closes https://github.com/manubot/rootstock/issues/440

`MANUBOT_USE_DOCKER=true` instructs Manubot to use Docker for PDF export. Defaults to `DOCKER_RUNNING` (whether Docker is running on the system).

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -11,12 +11,17 @@ export TZ=Etc/UTC
 # Default Python to read/write text files using UTF-8 encoding
 export LC_ALL=en_US.UTF-8
 
+
+# Set DOCKER_RUNNING to true if docker is running, otherwise false.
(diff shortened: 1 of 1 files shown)


dhimmel:
This enables running `MANUBOT_USE_DOCKER=false ./build/build.sh` to use WeasyPrint for PDF export on a system with Docker running.

`MANUBOT_USE_DOCKER=true ./build/build.sh` will try to use Docker regardless of whether its running, possibly causing an error. But that makes sense to me if the user has specified they'd like to use docker and it's not running.

multimeric:
Great! This sounds like exactly what I'm after, with a sensible default too.

agitter:
> Great! This sounds like exactly what I'm after, with a sensible default too.

👍  This variable name and default make sense to me too.  Thanks for the update @dhimmel.

PR #452, opened 2021-11-30
CI: use github.event context for DEFAULT_BRANCH

dhimmel:
Updates method added in https://github.com/manubot/rootstock/pull/399. See commit notes from https://github.com/manubot/rootstock/pull/399/commits/667d170eb2f1898156226456292d25c1c99377c4 on references for the old method. I believe `github.event.repository.default_branch` did not exist when we initially implemented this.

One benefit is this removes a bash command that does not work on macos due to an old grep. Running CI on macos or windows is helpful for prototyping our environment.

## References

- https://stackoverflow.com/a/68414395/4651668

### DIFF
--- a/.github/workflows/manubot.yaml
+++ b/.github/workflows/manubot.yaml
@@ -8,6 +8,8 @@ on:
     branches:
       - main
       - master
+  # NOTE: scheduled workflows aren't supported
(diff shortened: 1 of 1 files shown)


dhimmel:
It's possible this won't work for scheduled jobs, since scheduled jobs [are missing](https://github.community/t/github-event-repository-fork-context-does-detect-forks-in-scheduled-jobs/185925) the `github.event.repository.fork` context.

PR #454, opened 2021-12-11
CI: more workflow_dispatch options

dhimmel:
add SPELLCHECK & MANUBOT_USE_DOCKER options.

### DIFF
--- a/.github/workflows/manubot.yaml
+++ b/.github/workflows/manubot.yaml
@@ -24,14 +24,29 @@ on:
         type: boolean
         description: generate LaTeX output
         default: false
+      SPELLCHECK:
+        type: boolean
+        description: Check spelling
+        default: true
+      MANUBOT_USE_DOCKER:
+        type: boolean
+        description: Use Docker to generate PDF
+        default: true
 jobs:
   manubot:
     name: Manubot
     runs-on: ubuntu-latest
     env:
+      # Set Manubot preferences from workflow_dispatch input
+      # or a default value (for push/pull_request events).
+      # Edit default values by changing the true/false value after ||.
+      BUILD_PDF: ${{ format(github.event.inputs.BUILD_PDF) || true }}
+      BUILD_DOCX: ${{ format(github.event.inputs.BUILD_DOCX) || false }}
+      BUILD_LATEX: ${{ format(github.event.inputs.BUILD_LATEX) || false }}
+      SPELLCHECK: ${{ format(github.event.inputs.SPELLCHECK) || true }}
+      MANUBOT_USE_DOCKER: ${{ format(github.event.inputs.MANUBOT_USE_DOCKER) || true }}
+      # Set environment variables needed internally by Manubot.
       GITHUB_PULL_REQUEST_SHA: ${{ github.event.pull_request.head.sha }}
-      SPELLCHECK: true
-      MANUBOT_USE_DOCKER: true
     defaults:
       run:
         shell: bash --login {0}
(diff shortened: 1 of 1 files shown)


dhimmel:
Running the following workflow dispatch in https://github.com/dhimmel/manubot-rootstock/actions/runs/1567626264:

image

Here's what the env vars were in the logs:

```yaml
  env:
    BUILD_PDF: true
    BUILD_DOCX: false
    BUILD_LATEX: true
    SPELLCHECK: false
    MANUBOT_USE_DOCKER: false
    GITHUB_PULL_REQUEST_SHA: 
```

So it appears to have worked!

dhimmel:
One thing I don't like is that we have option defaults in 3 places:

1. GitHub Actions `on.workflow_dispatch.inputs` which only apply to workflow dispatch jobs
2. GitHub Actions `jobs.env` which apply to non-workflow dispatch jobs
3. `build.sh` defaults which apply to local jobs

Before this PR, `build.sh` would also set some defaults even when using GH actions: `SPELLCHECK` and `MANUBOT_USE_DOCKER`. `SPELLCHECK` needs to be accessible at the action stage since it requires extra dependencies.

@agitter do you have a sense how we should proceed?

agitter:
Nice addition, that looks great.

I agree that it is not ideal to have environment variable defaults set in so many places.  Even with the explanation in the comment, editing the values in `jobs.env` may also be slightly more intimidating now for someone who wants to edit the behavior of the GitHub Actions builds.

What restrictions do we have?  Do the `workflow_dispatch.inputs` defaults have to be hard-coded or could they be read from another source?  I don't have a suggestion in mind, but I'm thinking through whether any of these three places could defaults values set by the others (or in a new shared location?).  Even having the two workflow-related locations share defaults would be an improvement.

dhimmel:
> I'm thinking through whether any of these three places could defaults values set by the others (or in a new shared location?)

Yes a central location for defaults would be great, but I don't think the workflow_dispatch inputs can be templated or imported. The most immediate options I see are:

1. This PR where all variables are set to a default under  `jobs.manubot.env`
2. Setting SPELLCHECK in `jobs.manubot.env` and all the other variables in the "Setting Environment Variables (Workflow Dispatch)" step, such that they don't get set at all if its not a workflow_dispatch run.

I think perhaps I'm leaning towards 2?

agitter:
Yes, option 2 also makes sense to me.  We can't avoid setting defaults in `build.sh` because it can be run locally.  Setting defaults (those that are not needed until the build script runs) in `jobs.manubot.env` is redundant if the build script is going to set them anyway.

dhimmel:
> Yes, option 2 also makes sense to me

Okay switched to that in https://github.com/manubot/rootstock/pull/454/commits/dbb0952a72b52a07da238ccddaa79e87dac78771.

.github/workflows/manubot.yaml
```diff
@@ -47,6 +56,8 @@ jobs:
           echo "BUILD_PDF=${{ github.event.inputs.BUILD_PDF }}" >> $GITHUB_ENV
           echo "BUILD_DOCX=${{ github.event.inputs.BUILD_DOCX }}" >> $GITHUB_ENV
           echo "BUILD_LATEX=${{ github.event.inputs.BUILD_LATEX }}" >> $GITHUB_ENV
+          echo "SPELLCHECK=${{ github.event.inputs.SPELLCHECK }}" >> $GITHUB_ENV
```
dhimmel, line 59: Confirmed that this overwrites `jobs.manubot.env.SPELLCHECK` in https://github.com/dhimmel/manubot-rootstock/runs/4515234391?check_suite_focus=true#step:4:12

agitter (approved):

PR #456, opened 2022-01-25
Upgrade manubot & pandoc on 2022-01-24

dhimmel:
DataCite content negotiation is no longer used directly for DOIs. DOI metadata resolution should experience fewer failures, but certain fields might experience regressions.

### DIFF
--- a/build/environment.yml
+++ b/build/environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - ghp-import=2.0.2
   - jinja2=3.0.3
   - jsonschema=4.2.1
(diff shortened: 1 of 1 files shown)


dhimmel:
Did some local tests to ensure that the following metadata fields were operational:

```
manubot-randomize-author-order: true
manubot-randomize-author-order-seed: "set a custom seed here, otherwise use git commit hash"
```

These were added in https://github.com/manubot/manubot/pull/317.

agitter (approved):

PR #458, opened 2022-02-01
Update 00.front-matter.md

Adafede:
Hi,

While converting manuscripts to other formats through pandoc (for example ePub), I realized `<br>` causes some issues. Therefore, `\` is preferable over `<br>`. (see https://pandoc.org/MANUAL.html#extension-escaped_line_breaks)

I added some if statements to avoid incorrect linebreaks depending on author's information.

Also took the opportunity to add correspondence as default

Best,

### DIFF
--- a/content/00.front-matter.md
+++ b/content/00.front-matter.md
@@ -27,19 +27,26 @@ on {{manubot.date}}.
 
 {## Template for listing authors ##}
 {% for author in manubot.authors %}
-+ **{{author.name}}**<br>
++ **{{author.name}}** {% if author.correspondence is defined and author.correspondence == true -%}^[✉](#correspondence)^{%- endif -%}\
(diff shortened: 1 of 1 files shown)


dhimmel:
> I realized `<br>` causes some issues

Do you have a bit more details on the issues with `<br>`? It sounds like Pandoc is passing it though as HTML rather than parsing it as a hard line break?

Adafede:
Not so much, simply it breaks the ePub output. If you try 

```
# Pandoc --defaults for EPUB output.
# Load on top of common defaults.
to: epub
output-file: output/manuscript.epub
resource-path:
 - '.'
 - content
```
with `<br>`, the generated ePub will stop with an error when the `<br>` is reached. It does not with `\`. This does not impact the other outputs I tested...

PR #466, opened 2022-04-11
Adding pandoc-secnos to the build environment.

matthewturk:
In conjunction with https://github.com/manubot/manubot/pull/334 , this makes it possible to add section references to a manuscript.  I have not updated the github hash for manubot referenced in the build environment, but will if/when that pull request is accepted.

A sample build of this is now in the [yt paper build](https://yt-project.github.io/yt-4.0-paper/#selection-routines) where you can see the number of sections referenced.  I did note that all the sections are part of part 1, which I suspect might be an issue with the fact that I seem to be using `##` as the base section number.  That's probably fixable somehow on our end.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -76,6 +76,11 @@ For example, the following will override the figure number to be "S1" and set th
 
 We recommend always specifying the width of SVG images (even if just `width="100%"`), since otherwise SVGs may not render properly in the [WeasyPrint](https://weasyprint.org/) PDF export.
 
+### Sections
+
+Manubot supports section references, as provided by the [`pandoc-secnos`](https://github.com/tomduck/pandoc-secnos) library.
+Adding an anchor to a section header with `{#sec:section-name}` makes it available for citation using `@sec:section-name` inline in the text.
+
 ### Citations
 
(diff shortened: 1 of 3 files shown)


agitter (commented):
This also looks reasonable to me.  Thanks for updating the documentation.

It would make sense to merge https://github.com/manubot/manubot/pull/334 first so we can update the manubot version in the environment here.

dhimmel:
> It would make sense to merge https://github.com/manubot/manubot/pull/334 first so we can update the manubot version in the environment here

Yes, can upgrade manubot to manubot/manubot@fb3950460ea2ce2c5e17e7fcffeec1bb75b0d822

I'm in favor of adding pandoc-secnos if we think it's unlikely to break anything. For manuscripts without any `@sec:...` references, I'm hoping this filter will essentially do nothing.

Unfortunately, the pandoc-xnos suite hasn't been receiving frequent maintenance. So switching to [pandoc-crossref](https://github.com/lierdakil/pandoc-crossref) is probably the right move for Manubot. However, the syntax is similar, so merging this should not interfere with switching.

PR #470, opened 2022-07-07
Indicate Corresponding Author / Review Commons Compatibility

### LINKED ISSUE #469: Review Commons / Template Compatibility
I submitted my first manuscript to review commons using the standard manubot template. It came back with this comment:

> Corresponding author needs to be clearly indicated in the manuscript.
> Corresponding author marked in the manuscript needs to match the corresponding author marked on the submission page.

We might want to add a suggested corresponding author field + indicate in the template.

cgreene:
This PR aims to close #469.

### DIFF
--- a/content/00.front-matter.md
+++ b/content/00.front-matter.md
@@ -49,3 +49,12 @@ on {{manubot.date}}.
   {%- endif %}
   </small>
 {% endfor %}
+
+{% if manubot.authors|map(attribute='corresponding')|select|first %}
+  To whom correspondence should be addressed:
+  {% for author in manubot.authors %}
+    {% if author.corresponding is defined and author.corresponding is not none %}
+      {{author.initials}}: {{author.email}}
+    {% endif %}
+  {% endfor %}
+{% endif %}
--- a/content/00.front-matter.md
+++ b/content/00.front-matter.md
@@ -27,7 +27,9 @@ on {{manubot.date}}.
 
 {## Template for listing authors ##}
 {% for author in manubot.authors %}
-+ **{{author.name}}**<br>
++ **{{author.name}}**
+  {% if author.corresponding is defined and author.corresponding == true -%}^[✉](#correspondence)^{%- endif -%}
+  <br>
   {%- if author.orcid is defined and author.orcid is not none %}
     ![ORCID icon](images/orcid.svg){.inline_icon width=16 height=16}
     [{{author.orcid}}](https://orcid.org/{{author.orcid}})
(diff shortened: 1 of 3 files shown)


cgreene:
Ok, this build had corresponding set in the metadata.yaml file:
https://ci.appveyor.com/api/buildjobs/ovfe3iu2tyyvgs7n/artifacts/manuscript-1.0.283-8b1d9ee.pdf

This one does not:
https://ci.appveyor.com/api/buildjobs/98vk2kkhgisu57af/artifacts/manuscript-1.0.284-8526cd4.pdf

I didn't find a test suite, but hopefully this can suffice to demonstrate backwards compatibility. I'll add one more commit to put the flag back on in the rootstock so that the demo is complete.

content/00.front-matter.md
```diff
@@ -49,3 +49,12 @@ on {{manubot.date}}.
...
   </small>
 {% endfor %}
+
+{% if manubot.authors|map(attribute='corresponding')|select|first %}
+  To whom correspondence should be addressed:
+  {% for author in manubot.authors %}
+    {% if author.corresponding is defined and author.corresponding is not none %}
+      {{author.initials}}: {{author.email}}
+    {% endif %}
+  {% endfor %}
```
dhimmel, line 59: IIUC the indentation here is causing the corresponding author text to be treated as a code block:

> image

If we want a code block, I think the triple backtick syntax is more explicit. But a code block is probably the wrong formatting.

It might be worth looking through some of the existing corresponding author formats manuscripts have used (noted at https://github.com/manubot/rootstock/issues/469#issuecomment-1179199764) for inspiration on the best format for here.

dhimmel (commented):
Nice. Let's see what format @agitter thinks is best and then either @cgreene or I can update this PR.

content/00.front-matter.md
agitter: I agree that we probably want some other formatting besides a code block.

content/00.front-matter.md
```diff
@@ -49,3 +49,12 @@ on {{manubot.date}}.
   {%- endif %}
   </small>
 {% endfor %}
+
+{% if manubot.authors|map(attribute='corresponding')|select|first %}
+  To whom correspondence should be addressed:
```
agitter, line 54: This is where the corresponding symbol would be added.

agitter (commented):
I propose a hybrid of what is implemented here and what we had for this version of meta review:
- https://greenelab.github.io/meta-review/v/aa78ba9cbb2d874789cbfc9462dfb9aad19f655f/
- https://github.com/greenelab/meta-review/blob/aa78ba9cbb2d874789cbfc9462dfb9aad19f655f/content/00.front-matter.md?plain=1#L53-L56

We add a hard-coded symbol next to each corresponding author.  Then, we show that symbol in the conditional text block that lists the author contact information.

We should also update the "Manuscript metadata" section of USAGE before merging.

dhimmel:
as of https://github.com/manubot/rootstock/pull/470/commits/4f4f899de733fb7c5309623c80da626ef2f27e78, here is what it looks like with 1, 2, and 0 corresponding authors:

> image

> image

> image

Nice thing is that the names and emails can be copied directly into gmail and it's parsed correctly. Versions built on CI will link to GitHub Issues.

How does this format look? Also considering inline code around each author name + email. Could also use mailto links. I feel like the inline code might be slightly protected against spambots detecting the email address.

agitter:
> How does this format look?

That format looks very nice.  I think it looks better this way than with inline code, and it's unclear to me how much protection those provide against bots.

PR #476, opened 2022-10-21
github actions scheduled workflow

dhimmel:
`schedule` workflows now include the requisite `github.repository.event` context such that they can be used to build a manuscript alongside the other event types. See

- https://github.com/community/community/discussions/12269#discussioncomment-3747667
- https://github.com/dhimmel/dump-actions-context/actions/runs/3283984929/jobs/5409384642#step:2:41

### DIFF
--- a/.github/workflows/manubot.yaml
+++ b/.github/workflows/manubot.yaml
@@ -8,8 +8,10 @@ on:
     branches:
       - main
       - master
-  # NOTE: scheduled workflows aren't supported
-  # https://github.community/t/github-event-repository-fork-context-does-detect-forks-in-scheduled-jobs/185925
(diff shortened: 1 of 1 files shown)


agitter (approved):
That's great! This could be a nice feature for manuscripts that pull in external information and want to stay current.

Should we document this workflow possibility somewhere?  It may go unnoticed.

dhimmel:
> Should we document this workflow possibility somewhere? It may go unnoticed.

I think it's probably okay undocumented, as someone who wants to schedule the workflow will likely end up looking in `manubot.yaml` and seeing the commented out code. Expect this to be a somewhat niche feature.

Not against documentation, but will merge now.

PR #477, opened 2022-10-21
Fix author list when no author metadata is present

miltondp:
This PR adds a small change in front-matter to fix the author list when no author metadata (ORCID, Github, Twitter, etc) is given for an author. Below, I show an example of the current code and problem in [this manuscript](https://greenelab.github.io/phenoplier_manuscript/):

image

As you can see, when metadata is no present for an author, an unnecessary line break is added leaving an empty blank line.
This PR fixes this, producing this output:

image

### DIFF
--- a/content/00.front-matter.md
+++ b/content/00.front-matter.md
@@ -30,19 +30,23 @@ on {{manubot.date}}.
 + **{{author.name}}**
   {% if author.corresponding is defined and author.corresponding == true -%}^[✉](#correspondence)^{%- endif -%}
   <br>
-  {%- if author.orcid is defined and author.orcid is not none %} {% set line1 = true %}
+  {%- set has_ids = false %}
+  {%- if author.orcid is defined and author.orcid is not none %}
+  {%- set has_ids = true %}
     ![ORCID icon](images/orcid.svg){.inline_icon width=16 height=16}
     [{{author.orcid}}](https://orcid.org/{{author.orcid}})
   {%- endif %}
-  {%- if author.github is defined and author.github is not none %} {% set line1 = true %}
+  {%- if author.github is defined and author.github is not none %}
+    {%- set has_ids = true %}
     · ![GitHub icon](images/github.svg){.inline_icon width=16 height=16}
     [{{author.github}}](https://github.com/{{author.github}})
   {%- endif %}
-  {%- if author.twitter is defined and author.twitter is not none %} {% set line1 = true %}
+  {%- if author.twitter is defined and author.twitter is not none %}
+    {%- set has_ids = true %}
     · ![Twitter icon](images/twitter.svg){.inline_icon width=16 height=16}
     [{{author.twitter}}](https://twitter.com/{{author.twitter}})
   {%- endif %}
-  {%- if line1 is defined and line1 == true %}
+  {%- if has_ids == true %}
     <br>
   {%- endif %}
   <small>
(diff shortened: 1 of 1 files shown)


miltondp:
Another potential change is to add a new author in rootstock to add an example of this problem.

agitter:
First we can discuss the desired behavior and then look at the code changes.

We originally didn't account for this behavior because Manubot considers the ORCID to be [mandatory metadata](https://github.com/manubot/rootstock/blob/main/USAGE.md#manuscript-metadata).  However, we can't control what users actually do, so it probably is a good idea to be more robust to violations of that requirement.

>  add a new author in rootstock to add an example of this problem

I do think we would still want to strongly encourage ORCIDs, so we do not need to modify the example manuscript.

content/00.front-matter.md
```diff
@@ -30,18 +30,21 @@ on {{manubot.date}}.
 + **{{author.name}}**
   {% if author.corresponding is defined and author.corresponding == true -%}^[✉](#correspondence)^{%- endif -%}
   <br>
-  {%- if author.orcid is defined and author.orcid is not none %}
+  {%- if author.orcid is defined and author.orcid is not none %} {% set line1 = true %}
```
dhimmel, line 33: I wonder if the purpose would be a bit clearer if we:

`set has_ids = false` prior to the if statements and then had each if condition `set has_ids = true`. Then final check would be  `{%- if has_ids == true %}`.

Very similar so let's go with whichever @miltondp thinks will be most understandable for future people looking at the jinja code.

agitter: I like this proposed syntax change if it's okay with @miltondp.

miltondp: I think these are good changes, I'll add those.

miltondp:
I've added the suggested changes. Find below the differences in the author list with and without this patch.

output

I've added the diff to `metadata.yml` below in case you want to test it.

Maybe it doesn't matter, but just in case, John Doe and Jane Roe have [Github](https://github.com/johndoe) and/or [Twitter](https://twitter.com/johndoe) accounts.

```
diff --git a/content/metadata.yaml b/content/metadata.yaml
index d24da43..beacc34 100644
--- a/content/metadata.yaml
+++ b/content/metadata.yaml
@@ -16,6 +16,9 @@ authors:
       - Department of Something, University of Whatever
     funders:
       - Grant XXXXXXXX
+  - name: Another Author
+    affiliations:
+      - Department of Whatever, University of Something
   - github: janeroe
     name: Jane Roe
     initials: JR
```

content/00.front-matter.md
```diff
@@ -30,18 +30,25 @@ on {{manubot.date}}.
...
     · ![GitHub icon](images/github.svg){.inline_icon width=16 height=16}
     [{{author.github}}](https://github.com/{{author.github}})
   {%- endif %}
   {%- if author.twitter is defined and author.twitter is not none %}
+    {%- set has_ids = true %}
     · ![Twitter icon](images/twitter.svg){.inline_icon width=16 height=16}
     [{{author.twitter}}](https://twitter.com/{{author.twitter}})
-  {%- endif %}<br>
+  {%- endif %}
+  {%- if has_ids == true %}
```
agitter, line 49:
```suggestion
  {%- if has_ids %}
```
I haven't tested it, but I'm assuming this will work for a boolean

agitter (approved):
Looks good to me.  Thanks @miltondp.

My one comment is optional.  I'll leave this for @dhimmel to take another look before merging.

content/00.front-matter.md
dhimmel: I prefer `{%- if has_ids %}`, let's do that

miltondp: Done!

PR #478, opened 2022-11-12
Update Manubot on 2022-11-12 & date published support

dhimmel:


dhimmel:
Frontmatter is now like:

>  image

Not sure how to make it look better or more clear?

.github/workflows/manubot.yaml
```diff
@@ -89,7 +89,7 @@ jobs:
       - name: Build Manuscript
         run: bash build/build.sh
       - name: Upload Artifacts
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v3
```
dhimmel, line 92: I don't see artifacts showing up in GitHub actions anymore. Hoped this would fix it but didn't seem to.

agitter: They are still there.  From https://github.com/manubot/rootstock/actions you can pick a run, such as https://github.com/manubot/rootstock/actions/runs/3451537811.  Then the artifacts are at the bottom.  I believe this navigation changed at some point from how you used to find the artifacts.

content/00.front-matter.md
```diff
@@ -20,9 +20,11 @@ was automatically generated
 {% if manubot.ci_source is defined -%}
 from [{{manubot.ci_source.repo_slug}}@{{manubot.ci_source.commit | truncate(length=7, end='', leeway=0)}}](https://github.com/{{manubot.ci_source.repo_slug}}/tree/{{manubot.ci_source.commit}})
 {% endif -%}
-on {{manubot.date}}.
+on {{manubot.generated_date_long}}.
 </em></small>
 
+Published: {{manubot.date_long}}
```
agitter, line 26: I don't have too strong a preference of where we display this.  The main point is to show which template variable to use.  It might make sense to make it commented or include it as part of the existing commented block in lines 7-9.  Most manuscripts won't to display this date until a major version is finished.

dhimmel: I made the display conditional in https://github.com/manubot/rootstock/pull/478/commits/10f1616ef79d684dc2839cdc65b1d79fcb83a1fe... only display publication date if its different than the generated date.

Placing it with "A DOI-citeable version of this manuscript is available at..." but that text gets centered and it didn't quite look right to me.

.github/workflows/manubot.yaml
dhimmel: Ah yes, have to go to "Summary".

content/00.front-matter.md
agitter: 👍 that works for me

PR #497, opened 2023-06-29
Fix typo in table ref for 02.delete-me.md

rando2:
Space between #tbl: and name is not supported

I don't know if it would be worth adding a tip not to try to bold the Table name because I've been struggling with broken table refs for about thirty minutes and it turned out that was my mistake!

### DIFF
--- a/content/02.delete-me.md
+++ b/content/02.delete-me.md
@@ -215,7 +215,7 @@ Table: A table too wide to fit within page.
 | small    | black                            | white                |
 
 Table: A table with merged cells using the `attributes` plugin.
-{#tbl: merged-cells}
+{#tbl:merged-cells}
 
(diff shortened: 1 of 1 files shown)


dhimmel:
Nice surprised this bug existed for 4 years since 0dc5ea8c4530e8c1c0a56ac4ede48ac835c01251! Good find.

> I don't know if it would be worth adding a tip not to try to bold the Table name because I've been struggling with broken table refs for about thirty minutes and it turned out that was my mistake!

Can you elaborate? Does any bolding in the table name break the table ref? Do you have an example of a table ref with the issue?

rando2:
I will get a public repository up next week! It's the same one @agitter is requesting in manubot/catalog#70

PR #501, opened 2023-07-29
CI: install environment using mamba

dhimmel:
mamba is faster and has better error messages than conda

### DIFF
--- a/.github/workflows/manubot.yaml
+++ b/.github/workflows/manubot.yaml
@@ -84,7 +84,9 @@ jobs:
           activate-environment: manubot
           environment-file: build/environment.yml
           auto-activate-base: false
+          miniforge-variant: Mambaforge
           miniforge-version: 'latest'
+          use-mamba: true
       - name: Install Spellcheck
         if: env.SPELLCHECK == 'true'
         run: bash ci/install-spellcheck.sh


dhimmel:
Using mambda instead of conda on GitHub actions to install the environment reduced the install step from [2m 8s](https://github.com/manubot/rootstock/actions/runs/5579893752/job/15110661562) to [1m 27s](https://github.com/manubot/rootstock/actions/runs/5702454510/job/15454161347). Th 40 second reduction in build time will be nice.

dhimmel:
https://github.com/manubot/rootstock/pull/501/commits/ad49aa3da67e70ba51e90168294bc84374449d1a looks to decrease AppVeyor total job time from [3m 6s](https://ci.appveyor.com/project/manubot/rootstock/builds/46478073) to [2m 7s](https://ci.appveyor.com/project/manubot/rootstock/builds/47668516).

ci/install.sh
```diff
@@ -7,16 +7,16 @@
 set -o errexit \
     -o pipefail
 
-wget https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge-pypy3-$(uname)-$(uname -m).sh \
+wget https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-$(uname)-$(uname -m).sh \
```
dhimmel, line 10: Not sure why we were installing `Miniforge-pypy3` instead of `Miniforge3` before, probably just a mistake in https://github.com/manubot/rootstock/pull/450. `Miniforge-pypy3` includes `pypy` in addition to standard `python`, but I don't think we ever use `pypy`, so could save a bit of time just by switching to `Miniforge3`. But resolving the environment with `conda` would still be slow.

agitter (commented):
This is an improvement. I have had one environment where using conda with the libmamba solver gave a different set of packages than the default solver, but in general I still prefer mamba and agree with this change.

I thought when working with mamba we might need `mamba activate` instead of `conda activate` to activate the environment, but I see `conda activate` works in AppVeyor.

agitter (approved):

dhimmel:
> I have had one environment where using conda with the libmamba solver gave a different set of packages than the default solver

Good to know and something to keep in mind.

> I thought when working with mamba we might need `mamba activate` instead of `conda activate` to activate the environment, but I see `conda activate` works in AppVeyor.

Mamba docs used to suggest always using `conda activate`, but now they show `mamba activate`. For our needs probably simplest to stick with `conda activate` for a single command unless some issue arises.

PR #520, opened 2024-12-05
Remove explicitly set `mamba`-stuff from workflow file

### LINKED ISSUE #519: manuscript fails to build with "Warning: Future Miniforge releases will NOT build Mambaforge installers." error
A build of one of my [manuscripts](https://github.com/microct-ana-unibe-ch/microvasculature-manuscript) was hit by a 'brown-out' day specified in the ["Sunsetting Mambaforge" blog post](https://conda-forge.org/news/2024/07/29/sunsetting-mambaforge/) today.

I guess this is due to https://github.com/manubot/rootstock/blob/957ff8ebf04f1051aedb3fd17677e19ff2286b7a/.github/workflows/manubot.yaml#L82, but currently have no idea on how to fix it.

habi:
Removing the explicitly set `Mambaforge` variant and the call to `use-mamba` took care of an error preventing GitHub Actions to build a manuscript of mine during the so-called 'brownout days' stated in https://conda-forge.org/news/2024/07/29/sunsetting-mambaforge/

I might not see all the consequences of simply deleting those details, but submit this nonetheless as a PR :)

Fixes #519

### DIFF
--- a/.github/workflows/manubot.yaml
+++ b/.github/workflows/manubot.yaml
@@ -84,9 +84,6 @@ jobs:
           activate-environment: manubot
           environment-file: build/environment.yml
           auto-activate-base: false
-          miniforge-variant: Mambaforge
-          miniforge-version: 'latest'
-          use-mamba: true
       - name: Install Spellcheck
         if: env.SPELLCHECK == 'true'
         run: bash ci/install-spellcheck.sh


dhimmel (approved):
Thanks! This looks like the right solution to me. As long as mamba is found and can install the environment then we're all good.

PR #522, opened 2025-02-28
Updates ai-revision workflow for upcoming manubot-ai-editor multi-provider feature

falquaddoomi:
We'll shortly be introducing the ability for `manubot-ai-editor` to work with multiple providers (e.g., OpenAI, Anthropic, and more), facilitated by the [LangChain](https://python.langchain.com/docs/introduction/) library. These changes are included in PR manubot/manubot-ai-editor#71. That PR should be merged and a release created before we merge this PR, so this PR will remain a draft until that's completed.

Since we'll potentially be implementing many providers, we decided to allow a generic environment variable, `PROVIDER_API_KEY`, to specify the API key used for any one provider. While `manubot-ai-editor` could theoretically be run with multiple providers (each of which would need their own valid API key, of course), this particular workflow always runs the revision process with a single specific provider, so I assumed that using `PROVIDER_API_KEY` would be sufficient here. To not break things during the transition, the value will also be retrieved from `secrets.OPENAI_API_KEY` if `secrets.PROVIDER_API_KEY` is unavailable; this will likely be removed down the road.

I've also added `model_provider` to the inputs to specify the model provider (i.e., "openai" or "anthropic" for now). I'm unsure if we should provide the list of options as a dropdown and just update this whenever we add new providers, or keep it as a free-text string. I've included the list of providers we currently support for now, but I'm considering removing that in favor of either having a set list of options that we update, or directing the user to the manubot-ai-editor docs for an up-to-date list.

Finally, on @miltondp's advice, the default model engine is now `gpt-4o` rather than `gpt-4-turbo`.

### DIFF
--- a/.github/workflows/ai-revision.yaml
+++ b/.github/workflows/ai-revision.yaml
@@ -12,6 +12,16 @@ on:
         required: false
         type: string
         default: ''
+      model_provider:
+        description: 'Model Provider'
+        required: true
+        type: string
+        default: 'openai'
+      model_provider_email:
+        description: 'Model Provider Author Email'
+        required: true
+        type: string
+        default: '<support@openai.com>'
       model:
         description: 'Language model'
         required: true
@@ -56,7 +66,8 @@ jobs:
           pip install ${manubot_url}#egg=manubot[ai-rev]
       - name: Revise manuscript
         env:
-          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          PROVIDER_API_KEY: ${{ secrets.PROVIDER_API_KEY || secrets.OPENAI_API_KEY }}
+          AI_EDITOR_MODEL_PROVIDER: ${{ inputs.model_provider }}
           AI_EDITOR_LANGUAGE_MODEL: ${{ inputs.model }}
           AI_EDITOR_FILENAMES_TO_REVISE: ${{ inputs.file_names }}
           AI_EDITOR_CUSTOM_PROMPT: ${{ inputs.custom_prompt }}
@@ -66,9 +77,9 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v6
         with:
-          commit-message: 'revise using AI model\n\nUsing the OpenAI model ${{ inputs.model }}'
+          commit-message: 'revise using AI model\n\nUsing the ${{ inputs.model_provider }} model ${{ inputs.model }}'
           title: 'AI-based revision using ${{ inputs.model }}'
-          author: OpenAI model ${{ inputs.model }} <support@openai.com>
+          author: ${{ inputs.model_provider }} model ${{ inputs.model }} ${{ inputs.model_provider_email }}
           add-paths: |
             content/*.md
           branch: ${{ inputs.branch_name }}
--- a/.github/workflows/ai-revision.yaml
+++ b/.github/workflows/ai-revision.yaml
@@ -13,20 +13,15 @@ on:
         type: string
         default: ''
       model_provider:
-        description: 'Model Provider'
+        description: 'Model Provider (one of "openai", "anthropic")'
         required: true
         type: string
         default: 'openai'
-      model_provider_email:
-        description: 'Model Provider Author Email'
-        required: true
-        type: string
-        default: '<support@openai.com>'
       model:
         description: 'Language model'
         required: true
         type: string
-        default: 'gpt-4-turbo'
+        default: 'gpt-4o'
       custom_prompt:
         description: 'Custom prompt'
         required: false
@@ -79,7 +74,7 @@ jobs:
         with:
           commit-message: 'revise using AI model\n\nUsing the ${{ inputs.model_provider }} model ${{ inputs.model }}'
           title: 'AI-based revision using ${{ inputs.model }}'
-          author: ${{ inputs.model_provider }} model ${{ inputs.model }} ${{ inputs.model_provider_email }}
+          author: ${{ inputs.model_provider }} model ${{ inputs.model }} <support@${{ inputs.model_provider }}.com>
           add-paths: |
             content/*.md
           branch: ${{ inputs.branch_name }}


.github/workflows/ai-revision.yaml
```diff
@@ -66,9 +72,9 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v6
         with:
-          commit-message: 'revise using AI model\n\nUsing the OpenAI model ${{ inputs.model }}'
+          commit-message: 'revise using AI model\n\nUsing the ${{ inputs.model_provider }} model ${{ inputs.model }}'
           title: 'AI-based revision using ${{ inputs.model }}'
-          author: OpenAI model ${{ inputs.model }} <support@openai.com>
+          author: ${{ inputs.model_provider }} model ${{ inputs.model }} <support@${{ inputs.model_provider }}.com>
```
agitter, line 77: Will that support email address always work for future providers?

agitter (commented):
These generalization changes make sense and will be fast to approve. Let me know when you are ready to take it out of draft.

.github/workflows/ai-revision.yaml
falquaddoomi: Hey @agitter, thanks for the comment. Frankly, no, it won't always work; it's just chance that both OpenAI and Anthropic have support emails at a domain that can be easily constructed from the provider string. I also don't know how helpful these providers' support emails would be anyway.

I could see a few ways forward here:
- change it to an email address for the Manubot AI Editor maintainers (my preferred option, since we'd be in a position to help if someone did have a comment)
- add a workflow parameter that specifies the author email address (not my favorite option, since it's one more thing for the user to have to worry about)
- use `<${{ github.actor_id }}+${{ github.actor }}@users.noreply.github.com>` (the default for the [create-pull-request](https://github.com/peter-evans/create-pull-request) action) which would associate the commit with the user who ran the workflow

I'll talk to @miltondp, the owner of Manubot AI Editor about the above options and revise the PR accordingly.

miltondp: Thank you both for your work and comments. I'd go either with the first/preferred option or the third one. Or even not specify any email address if possible.

agitter: The first option seems less suitable to me because the Manubot AI Editor maintainers are the authors of the commit. Those maintainers shouldn't appear as contributing directly to many downstream manubot manuscripts. The third option is a reasonable default.

falquaddoomi: Thanks for the input, and that makes sense to me.

I've implemented option 3 as of commit [353d845](https://github.com/manubot/rootstock/pull/522/commits/353d8454bec9a9b5db6d1f083fe5a9562acd2f15). The commit author email is set to the same thing as the default value for the `author` field from the [create-pull-request action inputs](https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#action-inputs).

The display name remains `<model_provider> model <model_name>`; a example full author string would be:
`openai model gpt-4-turbo <312923+falquaddoomi@users.noreply.github.com>`.

I've verified that the GitHub user who submitted the workflow is listed as a co-author on the commit; you can see an example of that in my fork of rootstock here: https://github.com/falquaddoomi/rootstock/pull/1/commits/b8f36b0e8f69ce0d9ca2d770f6d0159d7ed44e0f.

agitter (approved):

falquaddoomi:
Summary of recent changes:
- In order to cut down on the number of options a user needs to select when dispatching the workflow, a few options can now be specified as org/repo variables, specifically:
  - `AI_EDITOR_MODEL_PROVIDER`: e.g., "openai" or "anthropic"; if "(repo default)" is selected as the model provider, this variable will be used to identify the model provider
  - `AI_EDITOR_LANGUAGE_MODEL`: the model from the specified provider to use; the user's choice always overrides this, but if they specify a falsy value (e.g, a blank string), it'll fall back to using this variable. If the variable isn't available, either, the default language model as defined in manubot-ai-editor for the provider will be used.
- The actual provider and model used are now scraped from manubot-ai-editor's output. I made this change mainly because the tool is the authoritative source on what is actually being used, but it's also is necessary when we defer the choice of the language model to manubot-ai-editor.

.github/workflows/ai-revision.yaml
```diff
@@ -26,7 +33,7 @@ on:
         description: 'Output branch'
         required: true
         type: string
-        default: 'ai-revision-gpt4turbo'
+        default: 'ai-revision-results'
```
miltondp, line 36: This is not high priority, but is it possible to use a variable here so that if we revise multiple times using a different provider/model, the output branch will not override the (potentially) existing one?

Again, not high priority.

.github/workflows/ai-revision.yaml
```diff
@@ -54,21 +61,50 @@ jobs:
           manubot_url=$(echo "$manubot_line" | awk -F"- " '{print $2}')
 
           pip install ${manubot_url}#egg=manubot[ai-rev]
+          
+          # install the multi-provider version of manubot-ai-editor over the current one, just for now
+          pip install git+https://github.com/manubot/manubot-ai-editor@langchain-anthropic-integration#egg=manubot-ai-editor
```
miltondp, line 66: Reminder to update this part.

.github/workflows/ai-revision.yaml
```diff
@@ -54,21 +61,50 @@ jobs:
...
+          if [[ "${AI_EDITOR_MODEL_PROVIDER}" = "(repo default)" ]]; then
+            # attempt to retrieve repo default model provider from the AI_EDITOR_MODEL_PROVIDER repo var
+            if [[ ! -z "${REPO_DEFAULT_MODEL_PROVIDER}" ]]; then
+              echo "* Using repo-default model provider AI_EDITOR_MODEL_PROVIDER repo variable: ${REPO_DEFAULT_MODEL_PROVIDER}"
+              export AI_EDITOR_MODEL_PROVIDER="${REPO_DEFAULT_MODEL_PROVIDER}"
+            else
+              echo "Warning: Repo-default provider was specified, but no AI_EDITOR_MODEL_PROVIDER repo variable was set, defaulting to 'openai'"
+              export AI_EDITOR_MODEL_PROVIDER="openai"
+            fi
+          fi
```
miltondp, line 87: Do we have documentation about all these variables?

falquaddoomi: We don't; I'll open a new PR in manubot-ai-editor to add that, thanks.

.github/workflows/ai-revision.yaml
falquaddoomi: I'm not sure, but that'd be a good idea; I'll look into it, thanks!

falquaddoomi: This is addressed in cc47cddd3241cbf24b879d4d942219693b36751a. The branch name field now defaults to `ai-revision-{1}`, where `{1}` is replaced with the final model name to match the older behavior. The provider could also be added to the branch name by specifying `{0}`.

.github/workflows/ai-revision.yaml
falquaddoomi: Removed as ccbc45f; it's now installed in the default way, via `manubot[ai-rev]`, which will pick up the latest `manubot-ai-editor` release from PyPI.

.github/workflows/ai-revision.yaml
falquaddoomi: As of 6b08f90ac300400ba5db02675038616b7d52f468, the new branch is named `ai-revision-{0}-{1}`, where `{0}` is the provider ("openai") and `{1}` is the model. I figured having more information in the branch name is better, especially since both values are short.

falquaddoomi:
Hi @agitter; I believe that this PR is ready for merging now. (@miltondp, mentioning you for visibility.)

Feel free to review it again if you like, since there have been some changes since the last time. I've tested it pretty extensively in my fork, but more testing can't hurt, I suppose. Thanks!

miltondp (approved):
I think this looks good, @falquaddoomi. I left a minor question/suggestion, if it makes sense. I would wait for @agitter for the final review and merge.

.github/workflows/ai-revision.yaml
```diff
@@ -56,20 +63,60 @@ jobs:
           pip install ${manubot_url}#egg=manubot[ai-rev]
       - name: Revise manuscript
         env:
-          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
-          AI_EDITOR_LANGUAGE_MODEL: ${{ inputs.model }}
+          PROVIDER_API_KEY: ${{ secrets.PROVIDER_API_KEY || secrets.OPENAI_API_KEY }}
```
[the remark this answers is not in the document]
falquaddoomi: Hey, thanks for the comment.  I'd originally added that to not break compatibility with users who hadn't set the new `PROVIDER_API_KEY` variable, but still had `OPENAI_API_KEY` set.

Now that I think about it, though, it's not necessary for it to check `OPENAI_API_KEY`, since `OPENAI_API_KEY` will take precedence over `PROVIDER_API_KEY` for OpenAI models, and if the user chooses an Anthropic model it'll fail earlier with a missing key rather than trying and failing to use an OpenAI `PROVIDER_API_KEY` with Anthropic later on in th eprocess.

I'll change it to just use `secrets.PROVIDER_API_KEY`; thanks again!

